PR #1, opened 2024-01-12
[FEQ] Shayan/feq-1425/ setup the package

shayan-deriv:
changes: 

- setup the project
- added Loader, Text and Button components


lib/components/Button/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+    type,
+    variant = 'contained',
+}) => {
+    const isContained = variant === 'contained';
+
+    const buttonClassNames = classNames(
+        'wallets-button',
+        `wallets-button__size--${size}`,
+        `wallets-button__variant--${variant}`,
+        `wallets-button__rounded--${rounded}`,
```
thisyahlen-deriv, line 47: Should be renaming this as well right 👀

shahzaib-deriv: Can we put this package out and as it's in beta we can put an improvement 
what do you say

shayan-deriv: @thisyahlen-deriv since we have both `wallets--` and `wallet--` prefixes I didn't do it in this PR. as @shahzaib-deriv said we can do this later. I just want to publish this and let u guys install and use it in deriv-app so we can make sure there won't be any building issues in deriv-app

aizad-deriv:
Just a suggestion maybe somewhere in the future there should be a root CSS Variable file for brand colors, status colors and even spacing. This was lacking from the previous components package from the Wallets. By implementing this not only the code will be cleaner and easier to maintained but it can also serve as a guide for us developers whenever we want to cross check  values on Figma and the ones we have 🙏🏼 
@shayan-deriv

lib/styles/wallet/styles/devices.scss
```diff
@@ -0,0 +1,23 @@
+/**
+ * Define Breakpoints here.
+ */
+$tablet-width: 768px;
+$desktop-width: 1024px;
```
aizad-deriv, line 5: can we also define the breakpoints inside of the root css and reference it here instead

lib/styles/shared/reset.scss
```diff
@@ -0,0 +1,147 @@
...
+    content: none;
+}
+
+table {
+    border-collapse: collapse;
+    border-spacing: 0;
+}
+
+strong {
+    font-weight: bold;
```
aizad-deriv, line 145: I always prefer values over the `bold`, `light` and `normal` syntax. It was an issue we have from the previous component library. 
```suggestion
    font-weight: 700;
```

lib/styles/shared/reset.scss
```diff
@@ -0,0 +1,147 @@
+/*------------------------------------*
```
aizad-deriv, line 1: I'm not sure if you are getting this from production but it looks good but do take a look at this maybe we can update it to the latest one 👉🏼  [here](https://elad2412.github.io/the-new-css-reset/)

package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@deriv/ui",
```
ali-hosseini-deriv, line 2:
```suggestion
  "name": "@deriv-com/ui",
```

ali-hosseini-deriv (changes requested):

markwylde-deriv (approved):

lib/components/Button/index.tsx
thisyahlen-deriv: Just scared this will be overlooked and we will never get back to this

shayan-deriv:
@Aizad-Deriv, appreciate your feedback. Currently, I'm generating various CSS files based on component usage, and I can't  figure out how to generate a global CSS variables file. Regarding the CSS normalizer, I've replaced the updated one and successfully built the package, but I need time to assess if it affects our existing style behaviors. 
I'll create cards for both your requests under the improvements card and address them later. 🙏

lib/components/Button/index.tsx
thisyahlen-deriv: can can

markwylde-deriv: @thisyahlen-deriv I think your concern is valid! I wonder if we start making use of the [Issues](https://github.com/deriv-com/ui/issues) tab to keep a track of this stuff.

thisyahlen-deriv: Yes @markwylde-deriv. What do you think @shayan-deriv

yashim-deriv:
Nice

ali-hosseini-deriv (approved):

aizad-deriv (approved):

thisyahlen-deriv (approved):

PR #6, opened 2024-01-15
refactor: renamed classname prefix from wallet to deriv

shayan-deriv:


lib/components/Button/Button.scss
```diff
@@ -33,7 +33,7 @@ $size-map: (
     ),
 );
 
-.wallets-button {
+.derivs-button {
```
shayan-deriv, line 36: for the plural ones I couldn't come up with better name. let's just keep it like this until we cleanup the sass files

PR #8, opened 2024-01-17
Aizad/FEQ-1480/Add Tooltip Component inside of Deriv UI

aizad-deriv:
- Created new Tooltip component for `@deriv/ui`
- Added different variants for Tooltip `general`, `dark` or `error`.
- Currently supports different `position` `variants` and `trigger-actions`.

image


lib/components/Tooltip/Tooltip.scss
```diff
@@ -0,0 +1,77 @@
+.wallets-tooltip {
```
shayan-deriv, line 1: lets rename `wallets` prefix to `derivs` and `wallet` to `deriv`(for now)

lib/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+    ) {
+      setActive(false);
+    }
+  };
+
+  useOnClickOutside(targetRef, (e) => handleClickOutside(e));
+
+  return (
+    <div
+      className="wallets-tooltip"
```
shayan-deriv, line 59: here u should rename the classname prefix as well 🙏

lib/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+      onMouseEnter={handleMouseEnter}
+      onMouseLeave={handleMouseLeave}
+      onClick={handleClick}
+      ref={targetRef}
+    >
+      {children}
+      {active && (
+        <div
+          className={`wallets-tooltip__content wallets-tooltip__content--${position} ${
```
shayan-deriv, line 68: same

lib/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+            className ?? ""
+          }`}
+        >
+          <div
+            className={`wallets-tooltip__arrow wallets-tooltip__arrow--${position}`}
```
shayan-deriv, line 73: same

lib/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+          />
+          <div className="wallets-tooltip__message">{message}</div>
```
shayan-deriv, line 75: same

lib/components/Tooltip/Tooltip.scss
```diff
@@ -0,0 +1,153 @@
...
+
+.deriv-arrow {
+  position: absolute;
+  width: 0;
+  height: 0;
+}
+
+// Position Classes
+
+.top {
```
sergei-deriv, line 54: @aizad-deriv I have doubts about these classes `.top`, `.left` and `.bottom`, can these styles overwrite the styles with the same name? Maybe would be better to add `.deriv-tooltip` prefix too? You can ignore this message if I'm totally wrong :)

sergei-deriv (approved):

lib/components/Tooltip/Tooltip.scss
aizad-deriv: You are absolutely right @sergei-deriv. I already made the changes thanks 🙏🏼

lib/hooks/useOnClickOutside.ts
```diff
@@ -0,0 +1,21 @@
+import { RefObject } from "react";
+import { useEventListener } from "./useEventListener";
```
thisyahlen-deriv, line 2: Question: Why do we need this when we can use the one in usehooks-ts 🤔

shayan-deriv: @thisyahlen-deriv later on if we decided to use more hooks from the library then we will use it directly from there

lib/hooks/useEventListener.ts
shayan-deriv, line 1: I think it would also be better to mention that this is from `usehooks-ts` in a comment at the top of the file

lib/hooks/useOnClickOutside.ts
thisyahlen-deriv: Not sure why we are reinventing the wheel here. But other than that LGTM 🙏

thisyahlen-deriv (approved):

shayan-deriv (approved):

lib/hooks/useOnClickOutside.ts
shayan-deriv: @thisyahlen-deriv I want to keep the bundle small and tidy.  I will check both ways and if the bundle size is not that much different then I'll include it

thisyahlen-deriv: Sure thing man

shayan-deriv (approved):

PR #9, opened 2024-01-18
[FEQ] shayan/feq-1452/added storybook+refactored components types

shayan-deriv:
changes: 

- setup storybook
- added stories for Text, Button, and Loader
- removed `React.FC<>` type from components


lib/components/Text/Text.scss
shayan-deriv, line 1: this is for fixing build warning

PR #13, opened 2024-01-19
Aizad/FEQ-1442/Input Component

aizad-deriv:
Created Input component with base stylings.
This input component can be extended to Text field, Password Field and Text Area.
It currently has these states: `active` & `disabled`.
It currently has these variants: `general`, `success` & `error`.
It comes with `HelperMessage` where based on the variant or the state of the input component it will change its color.
It has the option to pass text or icon on the right side of the input.
It is uncontrolled so you can use any third party libraries.   

https://github.com/deriv-com/ui/assets/103104395/c158e049-23f9-4ae9-b533-edfa5132a353

### DIFF
--- a/lib/components/Input/index.tsx
+++ b/lib/components/Input/index.tsx
@@ -1,0 +1,37 @@
+import React, { ComponentProps } from "react";
+import "./Input.scss";
+
+type InputProps = {
+  className?: string;
+  disabled?: boolean;
+  helpLineMessage?: string;
+  label?: string;
+  name?: ComponentProps<"input">["name"];
+  onChange?: ComponentProps<"input">["onChange"];
+  type?: ComponentProps<"input">["type"];
+  value?: ComponentProps<"input">["value"];
+};
+
+export const Input = ({
+  disabled = false,
+  label = "input",
+  name = "input",
+  onChange,
+  type = "text",
+  value,
+}: InputProps) => (
+  <div className="deriv-input-container">
+    <input
+      className="deriv-input-field"
+      type={type}
+      id={label}
+      value={value}
+      name={name}
+      onChange={onChange}
+      disabled={disabled}
+    />
+    <label className="deriv-input-label" htmlFor={label}>
+      <p className="deriv-input-label-text">{label}</p>
+    </label>
+  </div>
+);
--- a/lib/components/Input/index.tsx
+++ b/lib/components/Input/index.tsx
@@ -1,37 +1,39 @@
-import React, { ComponentProps } from "react";
+import React, { ChangeEvent, ComponentProps, useState } from "react";
+import clsx from "clsx";
 import "./Input.scss";
 
-type InputProps = {
-  className?: string;
-  disabled?: boolean;
-  helpLineMessage?: string;
+interface InputProps
+  extends Omit<ComponentProps<"input">, "style" | "placeholder"> {
   label?: string;
-  name?: ComponentProps<"input">["name"];
-  onChange?: ComponentProps<"input">["onChange"];
-  type?: ComponentProps<"input">["type"];
-  value?: ComponentProps<"input">["value"];
-};
+  helperMessage?: string;
+  error?: boolean;
+}
 
-export const Input = ({
-  disabled = false,
-  label = "input",
-  name = "input",
-  onChange,
-  type = "text",
-  value,
-}: InputProps) => (
-  <div className="deriv-input-container">
-    <input
-      className="deriv-input-field"
-      type={type}
-      id={label}
-      value={value}
-      name={name}
-      onChange={onChange}
-      disabled={disabled}
-    />
-    <label className="deriv-input-label" htmlFor={label}>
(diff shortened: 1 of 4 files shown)


lib/components/Input/index.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { ComponentProps } from "react";
+import "./Input.scss";
+
+type InputProps = {
+  className?: string;
+  disabled?: boolean;
+  helpLineMessage?: string;
```
shayan-deriv, line 7: not sure if the PR is still in progress but this is not used . I guess it's for the things like error/warning messages that goes under the input right? @aizad-deriv

lib/components/Input/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+};
+
+export const Input = ({
+  label,
+  id,
+  error,
+  message,
+  hasRightPlaceholder,
+  variant = "general",
+  ...rest
```
shayan-deriv, line 35:
```suggestion
  className,
  isHelperMessageDisabled,
  ...rest
```

lib/components/Input/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+}: InputProps) => {
+  return (
+    <div className="deriv-input">
+      <input
+        placeholder={label}
+        className={clsx(
+          "deriv-input--field",
+          { className: rest.className },
```
shayan-deriv, line 43:
```suggestion
          { className: className },
```

lib/components/Input/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+      {hasRightPlaceholder && (
+        <div className="deriv-input--right-content">{hasRightPlaceholder}</div>
+      )}
+      <div className="deriv-input--helper-message">
+        {message && (
+          <HelperMessage
+            message={message}
+            variant={variant}
+            error={error}
+            disabled={rest.disabled}
```
shayan-deriv, line 67:
```suggestion
            disabled={isHelperMessageDisabled}
```

shayan-deriv (commented):
I just added these suggestions, so we have them on the top suggestion list

lib/components/Input/index.tsx
shahzaib-deriv: why do we need `isHelperMessageDisabled`?

lib/components/Input/index.tsx
shahzaib-deriv: Helpermessage is relying on single disabled flag for the input. It is being used for setting the color to be synced

lib/components/Input/index.tsx
shahzaib-deriv, line 43:
```suggestion
          "deriv-input--field", rest.className,
```

lib/components/Input/index.tsx
aizad-deriv: I agree with @shahzaib-deriv the disabled message all depends on the input field same goes and so on.

shahzaib-deriv (changes requested):

lib/components/Input/index.tsx
aizad-deriv: resolved

lib/components/Input/index.tsx
aizad-deriv: In my opinion this doesn't bring any sort of value to the component because we are repeating over and over again. The helper message should behave the same as the input field.

lib/components/Input/index.tsx
aizad-deriv: Already resolved in the new commit 👍🏼

lib/components/Input/index.tsx
shayan-deriv: @aizad-deriv @shahzaib-deriv you're right. I just meant that since we were using it like `disabled={rest.disabled}` then we better get it out of the `rest` object. that's it.

lib/components/Input/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+          "deriv-input--label",
+          LabelVariant[error ? "error" : variant]
+        )}
+        htmlFor={id}
+      >
+        {label}
+      </label>
+      {hasRightPlaceholder && (
+        <div className="deriv-input--right-content">{hasRightPlaceholder}</div>
+      )}
```
shahzaib-deriv, line 62:
```suggestion
      {hasRightPlaceholder && (
        <div className="deriv-input--right-content">{hasRightPlaceholder}</div>
      )}
```

`hasRightPlaceholder` seems more like a boolean and IMO the content should be called only `rightPlaceholder` or something similar

lib/components/Input/index.tsx
shahzaib-deriv: @shayan-deriv Got your point. However using it from rest has no harm tho. Also @aizad-deriv already fixed it 👍

shahzaib-deriv (approved):

shayan-deriv (approved):

PR #25, opened 2024-01-26
[FEQ] aum/FEQ-1582/useDevice-hook

aum-deriv:


markwylde-deriv:
@aum-deriv what's the impacts of just adding usehooks-ts as a dependencies? It's [not that big](https://bundlephobia.com/package/usehooks-ts@2.10.0) and since it's typescript should only import what it needs.

aum-deriv:
> @aum-deriv what's the impacts of just adding usehooks-ts as a dependencies? It's [not that big](https://bundlephobia.com/package/usehooks-ts@2.10.0) and since it's typescript should only import what it needs.

Hi @markwylde-deriv,
I am not sure how the build process for TS works but we are trying to reduce the bundle size by directly copy-pasting the hooks for now.
@shayan-deriv Need your assistance to answer this query.

lib/hooks/useDevice.ts
```diff
@@ -0,0 +1,10 @@
+import { useMediaQuery } from "./useMediaQuery";
+
+export const useDevice = () => {
+    const isDesktop = useMediaQuery(`(min-width: 1024px)`);
+    const isMobile = useMediaQuery(`(max-width: 768px)`);
+    return {
+        isDesktop,
+        isMobile,
```
thisyahlen-deriv, line 8: Just a nit but can we expose all the variants like isSm, isMd, isLg like quill ?. Even if its not much its future proof even if we may not need it 🙏

matin-deriv (commented):
[Suggestion]: Add a short documentation for each of the hooks for as a quick reference!

lib/hooks/useDevice.ts
aum-deriv: @thisyahlen-deriv  think we should keep it simple for now and add variants as and when needed.

matin-deriv (approved):

lib/hooks/useMediaQuery.ts
shayan-deriv, line 1: since we are using more stuff from this hook now, I'll install the usehooks-ts package in a seperate PR.
cc: @thisyahlen-deriv

markwylde-deriv: Nice one Shayan! Love the process.

heorhi-deriv (approved):

PR #26, opened 2024-01-26
[FEQ]Yaswanth/FEQ-1439/Added Toggle Component

yaswanth-deriv:
Added a toggle component within the deriv/ui package, making it accessible for use across various packages by importing it directly from this library.

### DIFF
--- a/lib/components/ToggleSwitch/index.tsx
+++ b/lib/components/ToggleSwitch/index.tsx
@@ -1,0 +1,14 @@
+import React, { ChangeEvent, forwardRef } from 'react';
+import './ToggleSwitch.scss';
+
+interface ToggleSwitchProps {
+    onChange: (event: ChangeEvent<HTMLInputElement>) => void;
+    value: boolean;
+}
+
+export const ToggleSwitch = forwardRef<HTMLInputElement, ToggleSwitchProps>(({ onChange, value }, ref) => (
+    <label className='deriv-toggle-switch'>
+        <input checked={value} onChange={onChange} ref={ref} type='checkbox' />
+        <span className='deriv-toggle-switch__slider' />
+    </label>
+));
--- a/lib/components/ToggleSwitch/index.tsx
+++ b/lib/components/ToggleSwitch/index.tsx
@@ -1,14 +1,19 @@
 import React, { ChangeEvent, forwardRef } from 'react';
+import clsx from "clsx";
 import './ToggleSwitch.scss';
 
 interface ToggleSwitchProps {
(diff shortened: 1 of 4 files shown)


lib/components/ToggleSwitch/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+    onChange: (event: ChangeEvent<HTMLInputElement>) => void;
+    value: boolean;
+}
+
+export const ToggleSwitch = forwardRef<HTMLInputElement, ToggleSwitchProps>(({ onChange, value }, ref) => (
+    <label className='deriv-toggle-switch'>
+        <input checked={value} onChange={onChange} ref={ref} type='checkbox' />
+        <span className='deriv-toggle-switch__slider' />
+    </label>
+));
```
shayan-deriv, line 14: the reason is that the component itself should be able to internally handle the very basic behaviour which is toggling the value and if the user wants to add more stuff to it then they can pass onChange as well. we did the same thing for Tabs component as well.
what do you think @yaswanth-deriv 

```suggestion
export const ToggleSwitch = forwardRef<HTMLInputElement, ToggleSwitchProps>(({ onChange, value }, ref) => {
    const [checked, setIsChecked] = React.useState(value || false);
    const handleOnChange = (event: ChangeEvent<HTMLInputElement>) => {
        setIsChecked(!checked);
        onChange(event);
    }
 return (
    <label className='deriv-toggle-switch'>
        <input checked={value} onChange={handleOnChange} ref={ref} type='checkbox' />
        <span className='deriv-toggle-switch__slider' />
    </label>
  )
));
```

lib/components/ToggleSwitch/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { ChangeEvent, forwardRef } from 'react';
+import './ToggleSwitch.scss';
+
+interface ToggleSwitchProps {
+    onChange: (event: ChangeEvent<HTMLInputElement>) => void;
```
shayan-deriv, line 5:
```suggestion
    onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
```

lib/components/ToggleSwitch/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+    value: boolean;
```
shayan-deriv, line 6:
```suggestion
    value: boolean;
    wrapperClassName: string;
    className: string ,
    wrapperStyle: React.CSSProperties;
    style: React.CSSProperties;
```
and then use clsx()

shayan-deriv (approved):

PR #28, opened 2024-01-26
fix: imports from consuming projects

markwylde-deriv:
This fixes the imports so we can import components from the root, and still get the type definitions.

Example:

```tsx
import { Button } from '@deriv-com/ui'
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "private": false,
   "version": "0.0.1-beta.8",
   "type": "module",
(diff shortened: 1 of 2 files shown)


shayan-deriv (approved):
LGTM. thanks @markwylde-deriv 👍

shayan-deriv:
it's been merged in this PR:
https://github.com/deriv-com/ui/pull/41

PR #46, opened 2024-01-30
Aizad/FEQ-1560/Password Input Field

aizad-deriv:
- Added Password Input Field 
- Added Password Meter
- Added Scores and Messages
- Refactor Input Component and added new variants and styles
- Implement BEM class naming conventions within PasswordInput and Input

image


lib/components/PasswordInput/PasswordInput.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+  &-icon {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 100%;
+  }
+}
+
+.deriv-password-meter--bar {
```
wojciech-deriv, line 26: not sure if its a practice to follow BEM in deriv-com, but if it is, then better to use:
(I really like component styles to be scoped/isolated so there is no way to create dependencies between them)

.deriv-password {
    &__meter-bar { .. }
 }

package.json
```diff
@@ -50,6 +50,8 @@
     "@typescript-eslint/eslint-plugin": "^6.14.0",
     "@typescript-eslint/parser": "^6.14.0",
     "@vitejs/plugin-react": "^4.2.1",
+    "@zxcvbn-ts/core": "^3.0.4",
+    "@zxcvbn-ts/language-common": "^3.0.4",
```
wojciech-deriv, line 54: thats gonna be big, 
I'm wondering where should the lazy-loadijng happen, here in the UI library, or within other application using this library? 
Might be cool to give developers a component which is lazy-loaded by default,

wojciech-deriv: example of such component is within deriv-app: packages/wallets/src/components/Base/WalletPasswordFieldLazy/WalletPasswordFieldLazy.tsx

so whoever uses WalletPasswordFieldLazy, gets lazy-loaded version by default, which should be default shaken out of main bundle,

wojciech-deriv (approved):

hamza-deriv (approved):

yaswanth-deriv (approved):

package.json
aizad-deriv: I agree @wojciech-deriv but this requires some serious R&D, maybe in another PR we can address this issue and optimize it accordingly.

thisyahlen-deriv (approved):

shahzaib-deriv (approved):

shayan-deriv (approved):

PR #48, opened 2024-01-31
[FEQ] aum/FEQ-1654/pagelayout-improvements-remove-right-content-flex

aum-deriv:
#### Remove the right section prop
- Remove the right section prop and its corresponding rendering logic as it is decided that the right-hand side of the content will always be the part of the content. This is done to reduce the complexity for rendering the right-hand side as usually the content rendered in the content section has the control to change the behavior of the content on its right-hand side.
#### Fix: Content flex
- Applying `flex: 1` for the content section so that it can stretch to the entire width available to it.

https://github.com/deriv-com/ui/assets/125039206/102e21eb-1b46-47aa-a9bc-6c4568fb344d


src/stories/PageLayout.stories.tsx
```diff
@@ -78,8 +72,8 @@ const meta = {
         layout: "centered",
     },
     tags: ["autodocs"],
-    render: ({ children, left, right }) => (
-        <PageLayout left={left} right={right}>
+    render: ({ children, left }) => (
```
shayan-deriv, line 75: can we rename `left` to `sidebar`. it's more generic and even if we switch to RTL layout it still make sense.

aum-deriv: Hi @shayan-deriv,
Sure, will do this.

thisyahlen-deriv (approved):

lib/components/PageLayout/PageLayout.scss
```diff
@@ -8,18 +8,14 @@
         flex-direction: column;
     }
 
-    &__left {
+    &__sidebar {
         display: flex;
```
aizad-deriv, line 12: why are we declaring flex again within the container?

lib/components/PageLayout/PageLayout.scss
```diff
@@ -8,18 +8,14 @@
...
         flex-direction: column;
     }
 
     &__content {
         display: flex;
```
aizad-deriv, line 17: same here

shayan-deriv:
@aum-deriv need your opinion on this. how about instead of limit the layout to only have two columns we have sth more general where the consumer can pass as many columns as they need and control the style with props. 





for instance this:
```
<Layout>
    <Column flex={1} classname="sidebar">...</Column>
    <Column flex={2} classname="content">...</Column>    
</Layout>
```
Screenshot 2024-02-01 at 2 29 29 PM
```
<Layout>
    <Column flex={1} classname="sidebar">...</Column>
    <Column flex={2} classname="content">...</Column>    
    <Column flex={1} classname="content">...</Column>  
</Layout>
```
Screenshot 2024-02-01 at 2 29 46 PM

```
<Layout>
    <Column flex={1} classname="sidebar">...</Column>
    <Column flex={1} classname="content">...</Column>    
    <Column flex={1} classname="content">...</Column>  
</Layout>
```

Screenshot 2024-02-01 at 2 29 59 PM

then the consumer can decide which one is suitable for their case. for instance maybe the accounts team want to use context to handle the shared data between content and sideNote then they can go with the second example or the cashier decide to use the first one. 

what do you think guys?
@ali-hosseini-deriv @markwylde-deriv @aizad-deriv @amina-deriv @wojciech-deriv

wojciech-deriv (approved):

wojciech-deriv (approved):

wojciech-deriv:
> @aum-deriv need your opinion on this. how about instead of limit the layout to only have two columns we have sth more general where the consumer can pass as many columns as they need and control the style with props.
> 
> for instance this:
> 
> ```
> <Layout>
>     <Column flex={1} classname="sidebar">...</Column>
>     <Column flex={2} classname="content">...</Column>    
> </Layout>
> ```
> 
> Screenshot 2024-02-01 at 2 29 29 PM
> 
> ```
> <Layout>
>     <Column flex={1} classname="sidebar">...</Column>
>     <Column flex={2} classname="content">...</Column>    
>     <Column flex={1} classname="left-sidebar">...</Column>  
> </Layout>
> ```
> 
> Screenshot 2024-02-01 at 2 29 46 PM
> 
> ```
> <Layout>
>     <Column flex={1} classname="sidebar">...</Column>
>     <Column flex={1} classname="content">...</Column>    
>     <Column flex={1} classname="sidenote">...</Column>  
> </Layout>
> ```
> 
> Screenshot 2024-02-01 at 2 29 59 PM
> 
> then the consumer can decide which one is suitable for their case. for instance maybe the accounts team want to use context to handle the shared data between content and sideNote then they can go with the second example or the cashier decide to use the first one.
> 
> what do you think guys? @ali-hosseini-deriv @markwylde-deriv @aizad-deriv @amina-deriv @wojciech-deriv

My feeling is, that we shouldn't overly generalise. There are plenty of open-source layouting frameworks and I'm pretty sure we do not want to build another one internally due to overgeneralisation.
So basically - keep it specific to current deriv use cases. If right now we only need 2-column layout of specific widths, then lets keep it that way.
Or even generate two separate components - PageLayout2Col, PageLayout3Col - KISS / YAGNI.
I think its actually desired to not have full flexibility. I think its very desirable for ppl / designers to follow standard / agreed layouts across applications. To increase design consistency across products

aizad-deriv (approved):

lib/components/PageLayout/index.tsx
```diff
@@ -3,16 +3,14 @@ import { useDevice } from '../../hooks/useDevice';
...
+    sidebar?: JSX.Element;
 };
 
-export const PageLayout: React.FC<React.PropsWithChildren<PageLayoutProps>> = ({children, left, right}) => {
+export const PageLayout: React.FC<React.PropsWithChildren<PageLayoutProps>> = ({children, sidebar}) => {
     const {isMobile} = useDevice();
 
     return <div className="derivs-page-layout">
-        {left && !isMobile && <div className="derivs-page-layout__left">{left}</div>}
+        {sidebar && !isMobile && <div className="derivs-page-layout__sidebar">{sidebar}</div>}
```
shayan-deriv, line 13: do we need any styles for sidebar? because I see we haven't define anything in .scss file.

shayan-deriv (commented):
and one more thing. I noticed that we have a component called `Layout` as well. if we don't need that one then please remove it 🙏

lib/components/PageLayout/index.tsx
aum-deriv: Hi @shayan-deriv,
I don't think we need any. Please refer to [this](https://github.com/deriv-com/ui/pull/48#discussion_r1473828503)

shayan-deriv (approved):

PR #55, opened 2024-02-02
Aizad/FEQ-1660/Added Inline Message

aizad-deriv:
- Added Inline Message inside of UI
- Inline Message has variants: `info`, `success`, `warning` & `error`
- Inline Message has types: `filled`, `outlined`
- Accepts props: `className`, `variant`,`type`, `children` and `Icon`

image

### DIFF
--- a/lib/components/InlineMessage/VariantIcons.tsx
+++ b/lib/components/InlineMessage/VariantIcons.tsx
@@ -1,0 +1,58 @@
+export const ErrorIcon = () => (
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width="16"
+    height="16"
+    fill="none"
+    viewBox="0 0 16 16"
+  >
+    <g clipPath="url(#clip0_40941_3479)">
+      <path
+        fill="#EC3F3F"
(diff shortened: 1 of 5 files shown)


lib/components/InlineMessage/VariantIcons.tsx
```diff
@@ -0,0 +1,58 @@
+export const ErrorIcon = () => (
```
niloofar-deriv, line 1: Can we use these icons from quill-icons? 🤔

shayan-deriv: @niloofar-deriv we decided to skip using quill-icons for now. because we not not using that much icon . if the numbers gets high then we will definitely migrate to quill-icons 🙏

aizad-deriv: We can but we are only using a handful of it from quill-icons hence why as of now right now I am handling it like this.

thisyahlen-deriv (approved):

thisyahlen-deriv (approved):

shahzaib-deriv (approved):

sergei-deriv (approved):

shayan-deriv (approved):

PR #57, opened 2024-02-02
[FEQ]Yaswanth/FEQ-1448-Added Text Area Component

yaswanth-deriv:
Added a textarea component within the deriv/ui package, making it accessible for use across various packages by importing it directly from this library.
Screenshot 2024-02-02 at 4 49 56 PM

### DIFF
--- a/lib/components/TextArea/TextArea.scss
+++ b/lib/components/TextArea/TextArea.scss
@@ -1,0 +1,64 @@
+.deriv-textarea {
+    position: relative;
+    width: 100%;
+
+    & textarea {
+        resize: none;
+        width: 100%;
+        height: 9.6rem;
+        border-radius: 0.4rem;
+        border: 1px solid #d6dadb;
+        outline: none;
+        padding: 1rem 1.6rem;
+        font-size: 1.4rem;
+
+        &:focus {
+            border: 1px solid #85acb0;
+        }
+
+        &::placeholder {
+            color: #999999;
+        }
+        &:focus + label span {
+            color: #85acb0;
+        }
+        &:focus + label,
+        &[data-has-value='true'] + label {
+            transform: translate(0, -1.4rem) scale(0.75);
+            background-color: #ffffff;
+        }
+        &:has(~ label)::placeholder {
+            color: transparent;
+        }
+    }
+
+    & label {
+        display: block;
+        position: absolute;
+        left: 1.6rem;
+        top: 1rem;
+        transition: all 0.2s ease;
+        transform-origin: left top;
+        text-overflow: ellipsis;
+        overflow: hidden;
+    }
+
+    &--error {
+        & textarea,
+        & textarea:focus {
+            border: 1px solid #ec3f3f;
+            & + label span {
+                color: #ec3f3f;
+            }
+        }
+    }
+
+    &__footer {
+        display: flex;
+        line-height: 1.5;
+
+        & span {
+            margin-left: auto;
+        }
+    }
+}
--- a/lib/components/TextArea/index.tsx
+++ b/lib/components/TextArea/index.tsx
@@ -1,0 +1,67 @@
+import React, { HtmlHTMLAttributes, useState } from 'react';
+import clsx from 'clsx';
+import { Text } from "../Text";
+import { TGenericSizes } from '../../types';
+import './TextArea.scss';
+
+type TTextAreaProps = HtmlHTMLAttributes<HTMLTextAreaElement> & {
+    hint?: string;
+    isInvalid?: boolean;
+    label?: string;
+    maxLength?: number;
+    onInputChange: React.ChangeEventHandler<HTMLTextAreaElement>;
+    placeholder?: string;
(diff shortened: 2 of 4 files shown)


lib/components/TextArea/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const TextArea = ({
+    hint,
+    isInvalid = false,
+    label,
+    maxLength,
+    onInputChange,
+    placeholder,
+    textSize,
+    shouldShowCounter = false,
+    value,
```
thisyahlen-deriv, line 27: Lets expose all the props including className to <textarea

lib/components/TextArea/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+}: TTextAreaProps) => {
+    const [currentValue, setCurrentValue] = useState(value);
```
shahzaib-deriv, line 29: value shouldn't be kept inside the component

lib/components/TextArea/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+            )}
+            <div className='deriv-textarea__footer'>
+                {hint && (
+                    <Text as='p' color={isInvalid ? 'error' : 'less-prominent'} size={textSize}>
+                        {hint}
+                    </Text>
+                )}
+                {shouldShowCounter && (
+                    <Text color={isInvalid ? 'error' : 'less-prominent'} size={textSize}>
+                        {currentValue?.length || 0}/{maxLength}
```
shahzaib-deriv, line 61: to access the length, the value prop can be used directly to calculate the length rather than keeping the state inside this component

lib/components/TextArea/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+            })}
+        >
+            <textarea
+                data-has-value={!!currentValue}
+                onChange={event => {
+                    setCurrentValue(event.target.value);
+                    onInputChange?.(event);
+                }}
+                placeholder={placeholder}
+                value={currentValue}
```
shahzaib-deriv, line 44: we should expose each an every attribute available for textarea just like we did for the input component

shahzaib-deriv (changes requested):

lib/components/TextArea/TextArea.scss
```diff
@@ -0,0 +1,64 @@
...
+        text-overflow: ellipsis;
+        overflow: hidden;
+    }
+
+    &--error {
+        & textarea,
+        & textarea:focus {
+            border: 1px solid #ec3f3f;
+            & + label span {
+                color: #ec3f3f;
```
hasan-deriv, line 51: Is it possible to move all the color to css custom properties?

lib/components/TextArea/index.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { HtmlHTMLAttributes, useState } from 'react';
```
aizad-deriv, line 1: I'm not sure if it's doable but why is this component not extending from the existing `<Input/>` component? There are classNames and similar props with the same use cases. The only difference it has a bigger footprint.

lib/components/TextArea/TextArea.scss
```diff
@@ -0,0 +1,64 @@
+.deriv-textarea {
+    position: relative;
+    width: 100%;
+
+    & textarea {
+        resize: none;
+        width: 100%;
+        height: 9.6rem;
```
aizad-deriv, line 8: I'm not sure about using rem values. From my knowledge we haven' set the `1rem` to `10px`. If it is already configured you can just ignore if not then I suggest you to use to `px` instead.

lib/components/ToggleSwitch/index.tsx
```diff
@@ -3,12 +3,12 @@ import clsx from "clsx";
...
+    onChange: (event: ChangeEvent<HTMLInputElement>) => void;
     value: boolean;
-    wrapperClassName: string;
-    className: string;
-    wrapperStyle: React.CSSProperties;
-    style: React.CSSProperties;
+    wrapperClassName?: string;
+    className?: string;
+    wrapperStyle?: React.CSSProperties;
+    style?: React.CSSProperties;
```
aizad-deriv, line 11: if you want to take it to the next level you can always reference the Input component by:

```suggestion
    wrapperClassName?: ComponentProps<typeof Input>['wrapperClassName'];
    className?: ComponentProps<typeof Input>['className'];
    wrapperStyle?: React.CSSProperties;
    style?: React.CSSProperties;
```

yaswanth-deriv: Yes I can add this one..Thanks

lib/components/TextArea/TextArea.scss
yaswanth-deriv: Yes rem is configured..

lib/components/TextArea/TextArea.scss
shayan-deriv: @hasan-deriv we will refine the styles later

shayan-deriv (approved):

PR #60, opened 2024-02-05
Yaswanth/Added input props in dropdown component

yaswanth-deriv:
I needed to include the input component's props in the dropdown component and update the onChange prop to onInputChange.

### DIFF
--- a/lib/components/Dropdown/Dropdown.scss
+++ b/lib/components/Dropdown/Dropdown.scss
@@ -89,7 +89,7 @@
 
     &__items {
         position: absolute;
-        top: 2.7rem;
+        top: 2.9rem;
         width: 100%;
         display: flex;
         flex-direction: column;
--- a/lib/components/Dropdown/Dropdown.scss
+++ b/lib/components/Dropdown/Dropdown.scss
@@ -107,7 +107,9 @@
         & > :last-child {
             border-radius: 0 0 0.4rem 0.4rem;
         }
-
+        &--xs {
+            max-height: 12.5rem;
+        }
         &--sm {
             max-height: 22rem;
         }
--- a/lib/components/Dropdown/index.tsx
+++ b/lib/components/Dropdown/index.tsx
@@ -1,9 +1,9 @@
-import React, { isValidElement,HtmlHTMLAttributes ,useCallback, useEffect, useState } from 'react';
+import React, { isValidElement, HtmlHTMLAttributes, useCallback, useEffect, useState } from 'react';
 import clsx from 'clsx';
 import { useCombobox } from 'downshift';
 import { TGenericSizes } from "../../types";
 import { Text } from '../Text';
-import  {Input } from '../Input/index';
+import { Input } from '../Input/index';
 import './Dropdown.scss';
 
 type InputProps = React.ComponentProps<typeof Input>;
@@ -11,7 +11,6 @@ type TProps = HtmlHTMLAttributes<HTMLInputElement> & {
     disabled?: boolean;
     dropdownIcon: React.ReactNode;
     errorMessage?: InputProps['message'];
-    dropdownHeight?:string;
     icon?: React.ReactNode;
     isRequired?: boolean;
     label?: InputProps['label'];
@@ -19,7 +18,7 @@ type TProps = HtmlHTMLAttributes<HTMLInputElement> & {
         text?: React.ReactNode;
         value?: string;
     }[];
-    listHeight?: Extract<TGenericSizes, 'lg' | 'md' | 'sm'>;
+    listHeight?: Extract<TGenericSizes, 'lg' | 'md' | 'sm' | 'xs'>;
     name: InputProps['name'];
     onSearch?: (inputValue: string) => void;
     onSelect: (value: string) => void;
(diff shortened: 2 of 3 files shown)


thisyahlen-deriv (approved):

lib/components/Dropdown/index.tsx
```diff
@@ -7,7 +7,7 @@ import  {Input } from '../Input/index';
 import './Dropdown.scss';
 
 type InputProps = React.ComponentProps<typeof Input>;
-type TProps = {
+type TProps = HtmlHTMLAttributes<HTMLInputElement> & {
```
aizad-deriv, line 10: why are we not extending the types from Input component itself?

lib/components/Dropdown/index.tsx
```diff
@@ -20,12 +20,13 @@ type TProps = {
     }[];
     listHeight?: Extract<TGenericSizes, 'lg' | 'md' | 'sm'>;
     name: InputProps['name'];
-    onChange?: (inputValue: string) => void;
+    onInputChange?: (inputValue: string) => void;
```
aizad-deriv, line 23: I'm confused what's the difference between `onChange` and `onInputChange`?

lib/components/Dropdown/index.tsx
```diff
@@ -35,7 +36,7 @@ export const Dropdown = ({
     list,
     listHeight = 'md',
     name,
-    onChange,
+    onInputChange,
```
aizad-deriv, line 39: If `onInputChange` is a rename of a prop, `onChange` then you can do it like this:
```suggestion
    onChange: onInputChange,
```

lib/components/Dropdown/index.tsx
yaswanth-deriv: Yes.I will use the input component only.

lib/components/Dropdown/index.tsx
yaswanth-deriv: I got feedback that users found it confusing between onChange and onInputChange. Following the suggestion from @shayan-deriv, I am making this change.

shayan-deriv: we can have a better name like `onSearch`. I just wanted it to be more clear so ppl won't be confused

lib/components/Dropdown/index.tsx
```diff
@@ -134,24 +138,31 @@ export const Dropdown = ({
                     type='text'
                     value={value}
                     {...getInputProps()}
+                    {...rest}
                 />
             </div>
             <ul className={`deriv-dropdown__items deriv-dropdown__items--${listHeight}`} {...getMenuProps()}>
```
thisyahlen-deriv, line 144: Can you pass classname here as well

lib/components/Dropdown/index.tsx
```diff
@@ -20,25 +21,28 @@ type TProps = {
     }[];
     listHeight?: Extract<TGenericSizes, 'lg' | 'md' | 'sm'>;
```
thisyahlen-deriv, line 22: Can you also extract xs from TGenericSizes

lib/components/Dropdown/index.tsx
thisyahlen-deriv: And we have listHeight and dropdownHeight for the same thing?

lib/components/Dropdown/index.tsx
```diff
@@ -134,24 +138,31 @@ export const Dropdown = ({
                     type='text'
                     value={value}
                     {...getInputProps()}
+                    {...rest}
```
thisyahlen-deriv, line 141: ` leftPlaceholder={icon ?  icon : undefined}
                    rightPlaceholder={ (
                        <button
                            className={clsx('deriv-dropdown__button', {
                                'deriv-dropdown__button--active': isOpen,
                            })}
                        >
                            {dropdownIcon}
                        </button>
                    )}`
                    
                    can you extract this out and pass it as a component to avoid re-renders 😸

lib/components/Dropdown/Dropdown.scss
```diff
@@ -89,7 +89,7 @@
```
aizad-deriv, line 89: Most of the files currently are utilizing the `px` unit rather than `rem`. Let's keep it consistent throughout the whole project by using `px` as of now 👍🏼

lib/components/Dropdown/index.tsx
```diff
@@ -114,30 +129,24 @@ export const Dropdown = ({
         >
             <div className='deriv-dropdown__content'>
                 <Input
+                    className={className}
                     disabled={disabled}
```
thisyahlen-deriv, line 133: You dont need to specify these cause its already default component props. You already have {...rest} passed in the bottom, it will do the job for you

lib/components/Dropdown/index.tsx
```diff
@@ -18,28 +18,31 @@ type TProps = {
...
+    onSearch?: (inputValue: string) => void;
     onSelect: (value: string) => void;
     value?: InputProps['value'];
     variant?: 'comboBox' | 'prompt';
 };
 
+
 export const Dropdown = ({
+    className,
     disabled,
```
thisyahlen-deriv, line 32: Same here

lib/components/Dropdown/index.tsx
```diff
@@ -18,28 +18,31 @@ type TProps = {
...
         value?: string;
     }[];
-    listHeight?: Extract<TGenericSizes, 'lg' | 'md' | 'sm'>;
+    listHeight?: Extract<TGenericSizes, 'lg' | 'md' | 'sm' | 'xs'>;
     name: InputProps['name'];
-    onChange?: (inputValue: string) => void;
+    onSearch?: (inputValue: string) => void;
     onSelect: (value: string) => void;
     value?: InputProps['value'];
     variant?: 'comboBox' | 'prompt';
```
shayan-deriv, line 26: @yaswanth-deriv you should fix the variants as well.`prompt` variation is the one where you can type in it.

lib/components/Dropdown/Dropdown.scss
yaswanth-deriv: Yes.I am creating other PR to address these comments.Thanks

lib/components/Dropdown/index.tsx
yaswanth-deriv: Yes.I am creating other PR to address these comments.Thanks

PR #65, opened 2024-02-06
Shayan/webrel 2149/implement vertical tab

shayan-deriv:
changes: 
- added VerticalTab Component

image


lib/components/VerticalTab/VerticalTab.scss
```diff
@@ -0,0 +1,81 @@
+.vertical-tab__wrapper {
+    // background: #F2F3F4;
```
farrah-deriv, line 2: please remove this

hamza-deriv (approved):

lib/components/VerticalTab/VerticalTab.scss
```diff
@@ -0,0 +1,81 @@
...
+    padding: 8px;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    border-radius: 8px;
+}
+.shayan {
```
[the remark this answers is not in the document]
hamza-deriv: shayan flexing 👍

lib/components/VerticalTab/VerticalTabItems.tsx
```diff
@@ -0,0 +1,76 @@
...
+                                onSelectItemHandler={onSelectItemHandler}
+                                labelClassName={labelClassName}
+                                className={className}
+                                iconClassName={iconClassName}
+                            />
+                        )
+                    }
+                })}
+            </div>
+            <div className='vertical-tab__pane'>
```
shaheer-deriv, line 69: could please add spacing between the tab items and pane in styling?

lib/components/VerticalTab/VerticalTabItems.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { memo, useEffect, useState } from 'react';
+
+import { VerticalTabItem, type TabItem } from './VerticalTabItem'
+import { CollapsibleVerticalTabItem } from './CollapsibleVerticalTabItem';
+
+type VerticalTabItemsProps = {
+    activeTab: string;
```
farrah-deriv, line 7: what's the default tab if active tab is not passed?

lib/components/VerticalTab/CollapsibleVerticalTabItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+    iconClassName?: string;
+    labelClassName?: string;
+}
+
+
+const ArrowIcon = ({ is_open }: { is_open: boolean }) => <svg className={clsx(`vertical-tab__arrow`, {
+    'vertical-tab__arrow--open': is_open,
+})} xmlns="http://www.w3.org/2000/svg" width="17" height="16" viewBox="0 0 17 16" fill="none">
+    <path d="M8.66699 9.58579L13.9599 4.29289C14.3504 3.90237 14.9836 3.90237 15.3741 4.29289C15.7646 4.68342 15.7646 5.31658 15.3741 5.70711L9.3741 11.7071C8.98358 12.0976 8.35041 12.0976 7.95989 11.7071L1.95989 5.70711C1.56936 5.31658 1.56936 4.68342 1.95989 4.29289C2.35041 3.90237 2.98357 3.90237 3.3741 4.29289L8.66699 9.58579Z" fill="#333333" />
+</svg>;
```
ameerul-deriv, line 21: can we not put this in a separate file for this?

lib/components/VerticalTab/VerticalTab.tsx
```diff
@@ -0,0 +1,22 @@
+// Code:
```
ameerul-deriv, line 1: can we remove this 👀 
```suggestion
```

lib/components/VerticalTab/VerticalTabItem.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const VerticalTabItem = ({ tab, onClick, className, selectedTab, iconClassName,labelClassName }: VerticalTabItemProps) => {
+    return (
+        <div
+            className={
+                clsx(`vertical-tab__item`, {
+                    'vertical-tab__item--active': tab?.title === selectedTab
+                }, className)
+            }
+            onClick={() => onClick(tab?.title)}
```
ameerul-deriv, line 33: are the `?` necessary since it seems like it shouldn't be an optional prop passed to tab
```suggestion
                    'vertical-tab__item--active': tab.title === selectedTab
                }, className)
            }
            onClick={() => onClick(tab.title)}
```

lib/components/VerticalTab/VerticalTabItems.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    const onSelectItemHandler = (title: string) => {
+        const new_active_tab = findActiveTab(title)?.title;
+        setSelectedTab(() => new_active_tab ?? activeTab);
+        onSelectItem?.(title);
+    }
+
+
+    return (
+        <div className='shayan'>
```
ameerul-deriv, line 47: 👀

src/main.tsx
```diff
@@ -1,31 +1,58 @@
...
+    </g>
+    <defs>
+        <clipPath id="clip0_27043_25468">
+            <rect width="16" height="16" fill="white" />
+        </clipPath>
+    </defs>
+</svg>;
```
ameerul-deriv, line 16: can we move this icon to a separate file, same as the above

lib/components/VerticalTab/CollapsibleVerticalTabItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+    item: TabItem;
+    onSelectItemHandler: (title: string) => void;
+    selectedTab: string;
+    className?: string;
+    iconClassName?: string;
+    labelClassName?: string;
+}
+
+
+const ArrowIcon = ({ is_open }: { is_open: boolean }) => <svg className={clsx(`vertical-tab__arrow`, {
```
aizad-deriv, line 17: just a nit for this
```suggestion
const ArrowIcon = ({ is_open }: { is_open: boolean }) => <svg className={clsx('vertical-tab__arrow', {
```

lib/components/VerticalTab/CollapsibleVerticalTabItem.tsx
aizad-deriv: Maybe for now @shayan-deriv we can separate the icons in their own files. 
After this, we can create a new path for the `svgs` and import to the components that needs it.

lib/components/VerticalTab/CollapsibleVerticalTabItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    const onClickHandler = () => {
+        const shouldCollapse = !item?.subItems?.find((subItem) => subItem?.title === selectedTab);
+        if (shouldCollapse)
+            setOpen(!open);
+    }
+    return (
+        <div
+            key={item.title}
+            className={clsx(`collapsible-vertical-tab`)}>
```
aizad-deriv, line 41:
```suggestion
            className='collapsible-vertical-tab'>
```

lib/components/VerticalTab/CollapsibleVerticalTabItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <div className={clsx(`collapsible-vertical-tab__header`, {
```
aizad-deriv, line 42:
```suggestion
            <div className={clsx('collapsible-vertical-tab__header', {
```

src/main.tsx
aizad-deriv: @ameerul-deriv This is just a preview for us to see how the components work. But putting svgs in a seperate folder is valid 👍🏼

lib/components/VerticalTab/VerticalTabItems.tsx
aizad-deriv, line 69: 😭🙏🏻
```suggestion
            <div className='vertical-tab__panel'>
```

lib/components/VerticalTab/VerticalTab.scss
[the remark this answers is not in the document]
shayan-deriv: it was supposed to be temp 😄 removed it

lib/components/VerticalTab/VerticalTabItems.tsx
```diff
@@ -0,0 +1,129 @@
...
+ *          }
+ *      />
+ * </VerticalTab>
+ */
+
+export const VerticalTabItems = memo(({
+    items,
+    panelClassName,
+    wrapperClassName,
+    itemClassName,
```
shayan-deriv, line 71: `wrapperClassName` => green
`panelClassName` => pink
`itemClassName` => yellow

image

lib/components/VerticalTab/CollapsibleVerticalTabItem.tsx
shayan-deriv: @ameerul-deriv we might use quill-icons in the future. if not I'll keep all assets in seperate folder 🙏

lib/components/VerticalTab/VerticalTabItems.tsx
shayan-deriv: done 🙏 plus I pass classname for the wrapper so u can override the gap size between them. the default is now 25px

ameerul-deriv (approved):

lib/components/VerticalTab/CollapsibleVerticalTabItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+    className?: string;
+    iconClassName?: string;
+}
+
+
+const ArrowIcon = ({ is_open }: { is_open: boolean }) => <svg className={clsx('vertical-tab__arrow', {
+    'vertical-tab__arrow--open': is_open,
+})} xmlns="http://www.w3.org/2000/svg" width="17" height="16" viewBox="0 0 17 16" fill="none">
+    <path d="M8.66699 9.58579L13.9599 4.29289C14.3504 3.90237 14.9836 3.90237 15.3741 4.29289C15.7646 4.68342 15.7646 5.31658 15.3741 5.70711L9.3741 11.7071C8.98358 12.0976 8.35041 12.0976 7.95989 11.7071L1.95989 5.70711C1.56936 5.31658 1.56936 4.68342 1.95989 4.29289C2.35041 3.90237 2.98357 3.90237 3.3741 4.29289L8.66699 9.58579Z" fill="#333333" />
+</svg>;
```
aum-deriv, line 20: Hi @shayan-deriv,
Is this temporary? Will be having a local icons folder in this project or will we be using quill-icons?

lib/components/VerticalTab/VerticalTab.scss
aum-deriv, line 1: @shayan-deriv,
nit: Shouldn't we be using rem? As from my understanding, the elements will be able to scale with the base-font-size set by the project in which the component is used...

shaheer-deriv (approved):

hamza-deriv (approved):

lib/components/VerticalTab/VerticalTabItems.tsx
aum-deriv: @shayan-deriv,
I think its not a good idea to modify the styles of the children.
Please refer to this: https://deriv-group.slack.com/archives/C023SRL88VB/p1707287540236779
cc: @wojciech-deriv

lib/components/VerticalTab/VerticalTab.scss
aizad-deriv: @aum-deriv yes that's true but now with `v2` not all the projects are using `rem` for example: `tradershub-v2`. I think `px` is a safe unit and it works with those projects using `rem`

lib/components/VerticalTab/CollapsibleVerticalTabItem.tsx
aizad-deriv: It is temporary but atleast I feel we should put it in a separate file rather than just putting the svg right here.

lib/components/VerticalTab/VerticalTabItems.tsx
aizad-deriv: @aum-deriv there are pros and cons for not exposing the classname for each of the elements. However I do feel that having the flexibility to atleast change some part the styles for example, the padding of a container is fine. We are catering the needs for everyone and there is no one solution fits all.

aizad-deriv (approved):

lib/components/VerticalTab/VerticalTab.scss
```diff
@@ -0,0 +1,107 @@
+.vertical-tab__wrapper {
```
wojciech-deriv, line 1: minor: any reason for not putting this inside .vertical-tab ?

lib/components/VerticalTab/VerticalTab.scss
```diff
@@ -0,0 +1,107 @@
...
+    &__panel {
+        flex: 2;
+    }
+
+    &__arrow--open{
+        transform: rotate(180deg);
+    }
+}
+
+.collapsible-vertical-tab {
```
wojciech-deriv, line 71: that should  be in separate file for separate component

lib/components/VerticalTab/VerticalTab.scss
```diff
@@ -0,0 +1,107 @@
...
+        margin-right: 16px;
+    }
+
+    &__label {
+        flex:1;
+    }
+}
+
+
+.test-1{
```
wojciech-deriv, line 96: leftovers?
test-1 / test-2 / test-3 / test-4

lib/components/VerticalTab/VerticalTabItem.tsx
```diff
@@ -0,0 +1,42 @@
...
+    className?: string;
+    selectedTab: string;
+}
+
+export const VerticalTabItem = ({ tab, onClick, className, selectedTab }: VerticalTabItemProps) => {
+    return (
+        <div
+            className={
+                clsx('vertical-tab__item', {
+                    'vertical-tab__item--active': tab.title === selectedTab,
```
wojciech-deriv, line 30: is tab.title an arbitrary string, or is it a specific enum?
based on the naming itself, feels like one is title/translation and the other one is enum

lib/components/VerticalTab/VerticalTabItem.tsx
```diff
@@ -0,0 +1,42 @@
...
+                    'vertical-tab__item--disabled': tab.is_disabled
```
wojciech-deriv, line 31: hmm seems like the condition for --active and --disabled is different, 
does not feel good, it feels like it migth be possible to have both --disabled and --enabled class or none, 
ideallly, they both should use the same flag/variable

src/main.tsx
```diff
@@ -1,31 +1,58 @@
...
+  {
+    title: 'Item 3',
+    icon: Icon,
+  }
+]
 
+
+ReactDOM.createRoot(document.getElementById('root')!).render(
+  <React.StrictMode>
+    <div style={{margin:25, width:500}}>
```
wojciech-deriv, line 52: inline styles? not sure about standards in UI library but generally should be avoided (though I might be out of context for the UI library specifically)

lib/components/VerticalTab/VerticalTabItems.tsx
shayan-deriv: @aum-deriv I added it because previously for almost every component people asked us to support classname 😄 that's why I wanted to make it more flexible.

lib/components/VerticalTab/CollapsibleVerticalTabItem.tsx
shayan-deriv: in the future we might use quill-icon. otherwise I'll move them into asset folder

lib/components/VerticalTab/VerticalTab.scss
shayan-deriv: @aum-deriv we will address this concern later. for now we decided to only use px

lib/components/VerticalTab/VerticalTabItems.tsx
wojciech-deriv: Either way - probably better to just leave some default class and avoid adding extra props. 
If someone really want to style it, can do that via whateveer default classes are within component.

lib/components/VerticalTab/VerticalTab.scss
shayan-deriv: oops. removed it

lib/components/VerticalTab/VerticalTabItems.tsx
shayan-deriv: @wojciech-deriv you're right. for our case we want it not to be like headless-ui. but soon I'm going to cleanup the code and decide how we should handle the styling and standardize it .

src/main.tsx
shayan-deriv: this is just a playground for the developers who works on this project. soon I'll delete this since we have storybook now

aum-deriv (approved):

lib/components/VerticalTab/VerticalTab.scss
shayan-deriv: but in this case does it make any difference? because the consumer should use them all together. so that's why I put all styles in one file to avoid external request from consumer side

lib/components/VerticalTab/VerticalTabItem.tsx
shayan-deriv: it's string but not optional param. 
I didn't understand the issue. could u pleas elaborate more?

lib/components/VerticalTab/VerticalTabItem.tsx
shayan-deriv: but they are totally different. with active we only handle the style. but with disabled we also disable the functionality for onClick as well/ that's why I used two different things.

PR #70, opened 2024-02-09
[FEQ] Aum/FEQ-1699/ActionScreen-component

aum-deriv:
ActionScreen is a reusable component which is used in the cashier and accounts-settings pages for showing screens with design containing the following layout:
```
                                              icon
                                              title
                                           Description
                                               CTA (optional)
```
image

### DIFF
--- a/lib/components/ActionScreen/ActionScreen.scss
+++ b/lib/components/ActionScreen/ActionScreen.scss
@@ -1,0 +1,12 @@
+.derivs-action-screen {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    gap: 2.4rem;
+
+    &__info {
+        display: flex;
+        flex-direction: column;
+        gap: 0.8rem;
+    }
+}
--- a/lib/components/ActionScreen/index.tsx
+++ b/lib/components/ActionScreen/index.tsx
@@ -1,48 +1,57 @@
-import React, { ComponentProps, isValidElement, PropsWithChildren,  ReactNode } from 'react';
-import { Text } from '../Text';
-import './ActionScreen.scss';
+import React, {
+    ComponentProps,
+    isValidElement,
+    PropsWithChildren,
+    ReactNode,
+} from "react";
+import { Text } from "../Text";
+import "./ActionScreen.scss";
 
 type TProps = {
-    children?: ReactNode,
+    actionButtons?: JSX.Element;
+    children?: ReactNode;
     description?: ReactNode;
-    descriptionSize?: ComponentProps<typeof Text>['size'];
+    descriptionSize?: ComponentProps<typeof Text>["size"];
     icon?: ReactNode;
-    renderButtons?: () => ReactNode;
     title?: ReactNode;
-    titleSize?: ComponentProps<typeof Text>['size'];
+    titleSize?: ComponentProps<typeof Text>["size"];
 };
 
 /**
  * Generic component to display status / action screen / final screen.
- * (`Icon`, `Title`, `Description`, `CTA`)
+ * @param {TProps.actionButtons} actionButtons Button / group of buttons for CTA.
+ * @param {TProps.description} description Description content to be displayed as the description.
+ * @param {TProps.descriptionSize} descriptionSize Font-size of the description.
+ * @param {TProps.title} title Title content to be displayed.
+ * @param {TProps.titleSize} titleSize Font-size of the title.
  */
 export const ActionScreen: React.FC<PropsWithChildren<TProps>> = ({
+    actionButtons,
     children,
     description,
-    descriptionSize = 'md',
+    descriptionSize = "md",
     icon,
-    renderButtons,
     title,
-    titleSize = 'md',
+    titleSize = "md",
 }) => {
     return (
-        <div className='derivs-action-screen'>
+        <div className="derivs-action-screen">
             {icon}
-            <div className='derivs-action-screen__info'>
+            <div className="derivs-action-screen__info">
                 {title && (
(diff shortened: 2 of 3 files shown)


lib/components/ActionScreen/index.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { ComponentProps, isValidElement, PropsWithChildren,  ReactNode } from 'react';
+import { Text } from '../Text';
+import './ActionScreen.scss';
+
+type TProps = {
+    children?: ReactNode,
+    description?: ReactNode;
+    descriptionSize?: ComponentProps<typeof Text>['size'];
```
likhith-deriv, line 8: Rather than letting users to modify only size (`descriptionSize`, `titleSize`), why not accept classes to modify that section of the component?

lib/components/ActionScreen/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    titleSize?: ComponentProps<typeof Text>['size'];
+};
+
+/**
+ * Generic component to display status / action screen / final screen.
+ * (`Icon`, `Title`, `Description`, `CTA`)
+ */
+export const ActionScreen: React.FC<PropsWithChildren<TProps>> = ({
+    children,
+    description,
```
likhith-deriv, line 21: As I see the most dynamic section here is the `description`. So do we need to accept children prop separately? Based on the need, cant the users create custom sections under description and renderButton.
Can we structure this like`packages/account-v2/src/components/IconWithMessage/IconWithMessage.tsx`

lib/components/ActionScreen/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+                )}
+                {isValidElement(description) ? (
+                    description
+                ) : (
+                    <Text align='center' size={descriptionSize}>
+                        {description}
+                    </Text>
+                )}
+            </div>
+            {renderButtons?.()}
```
likhith-deriv, line 45: Its better to not enforce this section to be a function

lib/components/ActionScreen/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    return (
+        <div className='derivs-action-screen'>
+            {icon}
+            <div className='derivs-action-screen__info'>
+                {title && (
+                    <Text align='center' size={titleSize} weight='bold'>
+                        {title}
+                    </Text>
+                )}
+                {isValidElement(description) ? (
```
likhith-deriv, line 37: Nice function.. Will remember this 👍

lib/components/ActionScreen/index.tsx
aum-deriv: Hi @likhith-deriv,
IMHO, we should not be able to modify the styling of a component by passing classnames to it. Instead, we should try to cover all the variations that a component can have.

In case of this component, we just have a minor difference in the font-size of the title and description for one of the designs.

lib/components/ActionScreen/index.tsx
aum-deriv: Sorry, I couldn't get your point. Could you please elaborate on this 🙏

likhith-deriv: `title`, `description` props are React components. But `renderButton` is a function. So I asked is there a reason for doing this

lib/components/ActionScreen/index.tsx
aum-deriv: I am not sure about this as this is what was included in the traders-hub implementation as well.
Maybe traders-hub team can help with this 🙏 .

sergei-deriv: @aum-deriv @likhith-deriv We need children for the `ActionScreen`, because sometimes we need to put `formik` inside of the screen, for example. But we need buttons too, because it's more common case. Regarding `IconWithMessage` component, you have the same behaviour there, if children/button is not undefined then render it. The difference only in order of these components. Also, `children` is outside of the 2nd `div` element where description is. IMHO, we need to leave it as now, because any changes now will lead to style problems in many places

likhith-deriv: Cant the formik section come under the description prop itself?

aum-deriv: Yes @likhith-deriv, that makes sense.

image

sergei-deriv: @aum-deriv @likhith-deriv JFYI, just check these files:
- packages/tradershub/src/features/cfd/screens/ChangePassword/InvestorPassword/MT5ChangeInvestorPasswordInputsScreen.tsx
- packages/tradershub/src/components/SentEmailContent/SentEmailContent.tsx

These components uses `ActionScreen` with description and children... Pretty sure it can be easy modified, but as I said before it can lead to style issues...

But, anyway, I guess your solution is better when we have just the description, without children. Good job!

lib/components/ActionScreen/ActionScreen.scss
```diff
@@ -0,0 +1,12 @@
+.derivs-action-screen {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    gap: 2.4rem;
+
+    &__info {
+        display: flex;
+        flex-direction: column;
```
thisyahlen-deriv, line 9: You dont need to redeclare flex-col here, it will inherit from parent 😸

aum-deriv: Hi @thisyahlen-deriv,
Tried this but does not work...

shayan-deriv (approved):

PR #75, opened 2024-02-14
feat: added the table component

nada-deriv:
Table component using tanstack react/table + storybook
Screenshot 2024-02-14 at 5 07 53 PM

### DIFF
--- a/lib/components/Table/Table.scss
+++ b/lib/components/Table/Table.scss
@@ -1,0 +1,71 @@
+$general-color: #f2f3f4;
+$scroll-color: #d6dadb;
+$general-hover: #e6e9e9;
+
+@supports (-moz-appearance:none) {
+    .deriv-table__content {
+        scrollbar-color: transparent transparent;
+        scrollbar-width: thin;
+        /* Show scrollbar on hover */
+        &:hover {
+            scrollbar-width: thin; /* Show scrollbar on hover in Firefox */
+            scrollbar-color: $scroll-color transparent; /* Specify scrollbar color */
+        }
+    }
+
+}
+.deriv-table {
+    &__content {
+        overflow-y: auto;
+        display: flex;
+        flex-direction: column;
+        width: 100%;
+
+        @include mobile {
+            height: 100%;
+        }
+
+        &-row {
+            padding: 1.6rem;
+            &:hover {
+                background-color: $general-hover;
+            }
+            &:not(:last-child) {
+                border-bottom: 1px solid $general-color;
+            }
+        }
+    }
+
+    &__header {
+        display: grid;
+        border-bottom: 2px solid $general-color;
+        padding: 1.6rem;
+        grid-auto-flow: column;
+    }
+}
+
+::-webkit-scrollbar {
+    scrollbar-width: thin;
+    width: 5px;
+    height: 5px;
+    background-color: transparent;
+    border-radius: 10px;
+}
+
+::-webkit-scrollbar-track {
+    scrollbar-width: thin;
+    background-color: transparent;
+}
+
+::-webkit-scrollbar-thumb {
+    display: none;
+}
+
+:hover {
+    &::-webkit-scrollbar-thumb {
+        scrollbar-width: thin;
+        display: unset;
+        border-radius: 10px;
+        background-color: $scroll-color;
+    }
+}
--- a/lib/components/Table/Table.scss
+++ b/lib/components/Table/Table.scss
@@ -27,7 +27,7 @@ $general-hover: #e6e9e9;
         }
 
         &__row {
-            padding: 1.6rem;
+            padding: 16px;
             &:hover {
                 background-color: $general-hover;
             }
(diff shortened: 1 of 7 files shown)


lib/components/Table/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const topPosition = headerRef.current?.getBoundingClientRect().bottom ?? 0;
+
+    const { fetchMoreOnBottomReached } = useFetchMore({
+        loadMore: loadMoreFunction,
+        ref: tableContainerRef,
+        isFetching,
+    });
+
+    return (
+        <div className={clsx('w-full', tableClassname)}>
```
aizad-deriv, line 45: isn't this a Tailwind class

lib/components/Table/Table.scss
```diff
@@ -0,0 +1,71 @@
...
+        scrollbar-width: thin;
+        /* Show scrollbar on hover */
+        &:hover {
+            scrollbar-width: thin; /* Show scrollbar on hover in Firefox */
+            scrollbar-color: $scroll-color transparent; /* Specify scrollbar color */
+        }
+    }
+
+}
+.deriv-table {
```
aizad-deriv, line 17: Just a nit, please incorporate BEM className naming convention:

.deriv-table <== This is the block or the better know as the Parent.
.deriv-table__header, .deriv-table__content, .deriv-table__header <== These are the element.
.deriv-table__header--dark, .deriv-table__header--light <=== These are the modifiers or better know as Variants.

Having BEM naming convention helps in handling styling in different variants and maintain the consistency across the whole project.

lib/components/Table/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    renderHeader = () => <div />,
+    rowRender,
+    tableClassname = '',
+}: TProps<T>) => {
+    const table = useReactTable({
+        columns,
+        data,
+        getCoreRowModel: getCoreRowModel<T>(),
+        getGroupedRowModel: getGroupedRowModel<T>(),
+    });
```
aizad-deriv, line 32: In my opinion it is better to handle things such as ReactTable in the client side. For the components let's just handle the layout aswell as the styles.

shayan-deriv: @aizad-deriv how do you mean? so we should expose `useReactTable` to the consumer? 🤔 because from the consumer side all we need is just the `data` and based on that we should create the table layout with cols and rows. just like what we did in verticalTab.
could u please elaborate more ?

lib/components/Table/index.tsx
nada-deriv: oops.. forgot to remove it.. had it in p2p..

lib/components/Table/index.tsx
nada-deriv: @aizad-deriv  do you mean calling the usereacttable from our end and then passing the whole table as such to the deriv-com component? but isn't the current one a better approach?

lib/components/Table/Table.scss
```diff
@@ -25,7 +26,7 @@ $general-hover: #e6e9e9;
             height: 100%;
         }
 
-        &-row {
+        &__row {
             padding: 1.6rem;
```
aizad-deriv, line 30: let's not use rem units we are still using `px` across the whole project

lib/components/Table/index.tsx
aizad-deriv: I'm not sure how it works but I assume that we are passing the data somehow inside the table. My point is that the component of the table aswell as the rows and the columns should remain within the repo however the logic should be handle by the consumer because other teams might be using some other type of implementation.

aizad-deriv: That being said, if it's consistent throught the whole team then I think it should be fine you can just ignore this message 😂

lib/components/Table/Table.scss
```diff
@@ -0,0 +1,72 @@
...
+            &:not(:last-child) {
+                border-bottom: 1px solid $general-color;
+            }
+        }
+    }
+
+    &__header {
+        display: grid;
+        border-bottom: 2px solid $general-color;
+        padding: 1.6rem;
```
aizad-deriv, line 43: here aswell

lib/components/Table/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+                        </Text>
+                    ))}
+                </div>
+            )}
+            <div
+                className='deriv-table__content'
+                onScroll={e => fetchMoreOnBottomReached(e.target as HTMLDivElement)}
+                ref={tableContainerRef}
+                 //calculate height of the table content including the footer size.
+                 style={columns.length > 0 ? { height: `calc(100vh - ${topPosition}px - 3.6rem)` } : undefined}
```
aizad-deriv, line 60: here aswell

aizad-deriv (approved):

shayan-deriv (approved):

PR #79, opened 2024-02-16
[FEQ]Yaswanth/FEQ-1611-Add drop down with internal search box

yaswanth-deriv:
Added a dropdown with a search input inside, commonly used in p2p package for currency selection.

https://github.com/deriv-com/ui/assets/121096908/47e77809-5f81-4eb7-9085-ac3b813cb29e


ameerul-deriv:
hmmm the outcome looks different as expected, because the search icon is supposed to be to the left and not on the right, and there should be some spacing between the search and the dropdown above, not below the search input. 

and the label is still visible in the search input, which shouldn't be the case

lib/components/DropdownWithSearchBox/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+                    type="text"
+                    value={defVal}
+                    {...rest}
+                />
+            </div>
+            <ul className={`deriv-dropdown__items deriv-dropdown__items--${listHeight}`} {...getMenuProps()}>
+                {isOpen && (
+                    <div>
+                        <Input
+                            label={reactNodeToString(label)}
```
nada-deriv, line 169: we dont need the label for the search

yaswanth-deriv:
> hmmm the outcome looks different as expected, because the search icon is supposed to be to the left and not on the right, and there should be some spacing between the search and the dropdown above, not below the search input.
> 
> and the label is still visible in the search input, which shouldn't be the case

@ameerul-deriv  We can add search component for left or right by using rightplaceholder or leftplaceholder props

lib/components/DropdownWithSearchBox/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+            },
+            onInputValueChange({ inputValue }) {
+                setSearchInputValue(inputValue ?? '');
+                onSearch?.(inputValue ?? '');
+                if (shouldFilterListRef.current) {
+                    setItems(
+                        list.filter(item =>
+                            reactNodeToString(item.text)
+                                .toLowerCase()
+                                .includes(inputValue?.toLowerCase() ?? '')
```
nada-deriv, line 90: what about the case where the result is not found. shouldnt there be a prop for that? something like noResultsMessage.. "No results for ${text}"

yaswanth-deriv: we have helper message component which is included in Input component we can use that to show any error messages

ameerul-deriv: it is slightly different than the helper message, this is how it looks

Screenshot 2024-02-19 at 6 35 08 PM

yaswanth-deriv: Okay.I will add.Thanks

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        .deriv-textfield__field {
+            cursor: pointer;
+        }
+    }
+
+    &__field {
+        position: absolute;
+        inset: 0;
+        min-width: 0; /* this is required to reset input's default width */
+        padding-left: 2rem;
```
nada-deriv, line 40: i think these values are supposed to be in px instead

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
+.deriv-dropdown {
```
wojciech-deriv, line 1: naming - should it be in line with comonent name / file name for consistency and to avoid accidental overrides?

lib/components/DropdownWithSearchBox/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+import { useCombobox } from 'downshift';
+import { TGenericSizes } from '../../types';
+import { Text } from '../Text';
+import { Input } from '../Input/index';
+import './DropdownWithSearchBox.scss';
+
+type InputProps = React.ComponentProps<typeof Input>;
+type TProps = HtmlHTMLAttributes<HTMLInputElement> & {
+    disabled?: boolean;
+    dropdownIcon: React.ReactNode;
```
wojciech-deriv, line 12: wondering about API - shoduln't things like icons have default values? is there even a need to parametrize that? I would think that we actually do want to keep all the icons consistent across applications so probably even better to not allow it to be overriden.

Generally - YAGNI, less is more :)

lib/components/DropdownWithSearchBox/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+    dropdownHeight?: string;
+    icon?: React.ReactNode;
+    isRequired?: boolean;
+    label?: InputProps['label'];
+    list: {
+        text?: React.ReactNode;
+        value?: string;
+    }[];
+    listHeight?: Extract<TGenericSizes, 'lg' | 'md' | 'sm' | 'xs'>;
+    name: InputProps['name'];
```
wojciech-deriv, line 24: whats the purspose of allowing to override it? onSearch, onSelect should be enough, we shoulnd't allow ppl to override internals of dropddown without solid justification. Best to keep it as simple as possible, and only add necessary props

yaswanth-deriv:
We are merging this PR..I will create subtasks for your comments and resolve it in separate PR.Thanks

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
shayan-deriv: I agree. @yaswanth-deriv would u please update it? because we already have dropdown and it might have conflicts.

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
shayan-deriv, line 40:
```suggestion
        padding-left: 20px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        pointer-events: none;
+
+        & label {
+            color: var(--system-light-5-active-background, #999);
+        }
+    }
+
+    &__button {
+        all: unset;
+        right: 1.6rem;
```
shayan-deriv, line 16:
```suggestion
        right: 16px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+    &__field {
+        position: absolute;
+        inset: 0;
+        min-width: 0; /* this is required to reset input's default width */
+        padding-left: 2rem;
+        display: flex;
+        flex-grow: 1;
+        font-family: inherit;
+        outline: 0;
+        font-size: 1.4rem;
```
shayan-deriv, line 45:
```suggestion
        font-size: 14px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+            background-color: transparent;
+        }
+
+        &::placeholder {
+            color: transparent;
+        }
+    }
+
+    &__field:placeholder-shown ~ &__label {
+        font-size: 1.4rem;
```
shayan-deriv, line 61:
```suggestion
        font-size: 14px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        cursor: text;
+        top: 30%;
+        padding: 0;
+    }
+
+    &__field:placeholder-shown ~ &__label--with-icon {
+        left: 4.4rem;
```
shayan-deriv, line 68:
```suggestion
        left: 44px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+
+    &__items {
+        position: absolute;
+        top: 45px;
+        width: 100%;
+        display: flex;
+        flex-direction: column;
+        align-items: flex-start;
+        z-index: 2;
+        border-radius: 0.4rem;
```
shayan-deriv, line 98:
```suggestion
        border-radius: 4px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        background: var(--system-light-8-primary-background, #fff);
+        box-shadow: 0 3.2rem 6.4rem 0 rgba(14, 14, 14, 0.14);
```
shayan-deriv, line 100:
```suggestion
        box-shadow: 0 32px 64px 0 rgba(14, 14, 14, 0.14);
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        overflow-y: auto;
+
+        & > :first-child {
+            border-radius: 0.4rem 0.4rem 0 0;
```
shayan-deriv, line 104:
```suggestion
            border-radius: 4px 4px 0 0;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        & > :last-child {
+            border-radius: 0 0 0.4rem 0.4rem;
```
shayan-deriv, line 108:
```suggestion
            border-radius: 0 0 4px 4px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        &--sm {
+            max-height: 22rem;
```
shayan-deriv, line 112:
```suggestion
            max-height: 220px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        &--md {
+            max-height: 42rem;
```
shayan-deriv, line 116:
```suggestion
            max-height: 420px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        &--lg {
+            max-height: 66rem;
```
shayan-deriv, line 120:
```suggestion
            max-height: 660px;
```

lib/components/DropdownWithSearchBox/DropdownWithSearchBox.scss
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    &__icon {
+        position: absolute;
+        left: 1.6rem;
+        width: 1.6rem;
+        height: 1.6rem;
```
shayan-deriv, line 128:
```suggestion
        left: 16px;
        width: 16px;
        height: 16px;
```

lib/components/DropdownWithSearchBox/index.tsx
shayan-deriv: @wojciech-deriv previously we had them internally in the component but it turned out that in some cases we also can have other icons for it. that's why we made parametrize it but I agree that we should have a default icon.

@yaswanth-deriv would you please add a default icon for search and also for cancelIcon? 

just use a normal svg because later we probably replace them with quill-icons.

```
const CancelIcon = (
    <svg
        xmlns="http://www.w3.org/2000/svg"
        viewBox="0 0 24 24"
        width="10"
        height="10"
        fill="none"
        stroke="black"
        stroke-width="4"
        stroke-linecap="round"
        stroke-linejoin="round"
    >
        <path d="M18 6L6 18M6 6l12 12" />
    </svg>
```

shayan-deriv: @yaswanth-deriv actually based on the design file we don't need to pass the icons for this component as a prop.
please remove cancelIcon, dropdownIcon and also icon 🙏 

https://www.figma.com/file/K12B4QhJM4GllKn2CLGYzv/Cross-border-ads?node-id=19425%3A68459&mode=dev

PR #83, opened 2024-02-19
george / add top alignment for message icon

heorhi-deriv:
Wrong (current) behaviour: 
Screenshot 2024-02-19 at 15 18 22
Proper alignment according `Figma` design:
Screenshot 2024-02-19 at 15 20 41


wojciech-deriv (approved):

shayan-deriv (approved):

shayan-deriv:
@heorhi-deriv please resolve the conflict 🙏

aizad-deriv:
Looks good @heorhi-deriv but what happens if the message is single line?  it should be aligned in the center. Having it at the top works but for single-line it might look abit off. I suggest to add a prop for `iconPosition` which the client can set either at the top or the center.

Regardless LGTM.

heorhi-deriv:
> Looks good @heorhi-deriv but what happens if the message is single line? it should be aligned in the center. Having it at the top works but for single-line it might look abit off. I suggest to add a prop for `iconPosition` which the client can set either at the top or the center.
> 
> Regardless LGTM.

Ok, @aizad-deriv i see your concern. Ok let me pass the prop

niloofar-deriv (approved):

lib/components/InlineMessage/index.tsx
```diff
@@ -57,7 +59,9 @@ export const InlineMessage = ({
         )}
     >
         {(variant !== "general" || icon) && (
-            <div className="deriv-inline-message__icon">
+            <div
+                className={`deriv-inline-message__icon deriv-inline-message__icon--${iconPosition}`}
```
aizad-deriv, line 63: can we use `clsx` for this one like this:
```suggestion
                className={clsx('deriv-inline-message__icon,'  {
                'deriv-inline-message__icon--top' : top,
                'deriv-inline-message__icon--center:  center,
                'deriv-inline-message__icon--bottom: bottom
                }
```
easier to see and debug aswell if you do it like this.

heorhi-deriv: @aizad-deriv sorted

aizad-deriv (approved):

niloofar-deriv (approved):

PR #84, opened 2024-02-19
[FEQ]Yaswanth/FEQ-1752/Added isFullWidth prop 

yaswanth-deriv:
Need to add isFullWidth prop in dropdown component to make changes in width of the dropdown


lib/components/Dropdown/Dropdown.scss
```diff
@@ -99,6 +98,11 @@
         background: var(--system-light-8-primary-background, #fff);
         box-shadow: 0 32px 64px 0 rgba(14, 14, 14, 0.14);
         overflow-y: auto;
+        width: 328px;
+
+        &--full {
```
wojciech-deriv, line 103: question: should it be "full" by default?

whats the general strategy for component sizing - should they be flexible and deriv-com/ui clients should set the sizes, or should they have widths defined by deriv-com/ui?

yaswanth-deriv: I am using the same format used in Input component, which is also applied to the dropdown. Thus, I've been consistent in following these guidelines

wojciech-deriv: I do not mind if thats the standard with the whole library, though, IMO all the components should be 100% width by default, to accomodate for variety of responsive layouts, 

not ideal if they have fixed widths as that means, that every application with responsive layout, have to override default width with responsive width and add extra .css for that,

wojciech-deriv (approved):

lib/components/Dropdown/Dropdown.scss
yaswanth-deriv: Certainly, what you mentioned is correct. I'll consult with my team members and ensure that we stick to 100% as the standard, making any necessary changes accordingly

jim-deriv (approved):

lib/components/Dropdown/index.tsx
```diff
@@ -155,11 +157,8 @@ export const Dropdown = ({
                     {...rest}
                 />
             </div>
-            <ul
-                className={`deriv-dropdown__items deriv-dropdown__items--${listHeight}`}
-                {...getMenuProps()}
-            >
-                {isOpen &&
+            <ul className={`deriv-dropdown__items deriv-dropdown__items--${listHeight} deriv-dropdown__items--${isFullWidth ? "full" : isFullWidth}`} {...getMenuProps()}>
```
ameerul-deriv, line 160: can we not write this using clsx library?

yaswanth-deriv: Yes I will use clsx then

ameerul-deriv (approved):

lib/components/Dropdown/Dropdown.scss
shayan-deriv: @yaswanth-deriv I think for this case we should have full width by default. because as I checked when we open the dropdown, the items container are full-width. so in that case then the dropdown selector should be full width.

shayan-deriv: @wojciech-deriv u're right. we should stick with one of the approaches for all components. I'll check them and create card for them so we can make it consistent through all components.

shayan-deriv (approved):

PR #85, opened 2024-02-20
Aizad/ Fix Checkbox id returning errors inside of deriv.app

aizad-deriv:
- remove useId() from Checkbox component.
- make `name` as a required prop.
- use `name` instead as a fallback for the `id`.
- updated storybook.

image

### DIFF
--- a/lib/components/Checkbox/index.tsx
+++ b/lib/components/Checkbox/index.tsx
@@ -1,9 +1,13 @@
-import { ComponentProps, ReactNode, Ref, forwardRef, useId } from "react";
+import { ComponentProps, ReactNode, Ref, forwardRef } from "react";
 import clsx from "clsx";
 import "./Checkbox.scss";
 
 interface CheckboxProps
-  extends Omit<ComponentProps<"input">, "placeholder" | "defaultChecked"> {
+  extends Omit<
+    ComponentProps<"input">,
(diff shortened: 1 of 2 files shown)


thisyahlen-deriv (approved):

shahzaib-deriv (approved):

lib/components/Checkbox/index.tsx
```diff
@@ -20,17 +24,16 @@ export const Checkbox = forwardRef(
...
     ref: Ref<HTMLInputElement>
   ) => {
-    const id = useId();
-
     return (
       <div className={clsx("deriv-checkbox", wrapperClassName)}>
         <div className="deriv-checkbox__wrapper">
           <input
-            id={rest.id ?? id}
+            id={rest.id ?? name}
```
yashim-deriv, line 36:
```suggestion
            id={rest.id ?? (name + Date.now().toString())}
```
Since useId requires React 18, and you need a unique identifier, you can append the component name to Date.now() to get some amount of uniqueness from the id

ali-hosseini-deriv (approved):

lib/components/Checkbox/index.tsx
aizad-deriv: Willl implement this in the next fix then 😂

PR #97, opened 2024-02-29
Aizad/ Update scss file structure

aizad-deriv:
- Restructuring the current `SCSS` files following industry practices.
- Removing unused styles and libraries
- Re-adding all the root values and colors following current design implementation.
- Supporting both light and dark color scheme based on current design implementation.
- One source of Truth for all styles definitions


lib/styles/wallet/styles/devices.scss
```diff
@@ -1,6 +1,3 @@
-/**
- * Define Breakpoints here.
- */
 $tablet-width: 768px;
 $desktop-width: 1024px;
```
shayan-deriv, line 2: shouldn't we update this with the latest values confirmed by the designers? 
```
600px; // max-width for mobile
1280px; // min-width for desktop
// for tablet the range is in between
```

lib/styles/abstracts/_devices.scss
```diff
@@ -0,0 +1,62 @@
+/**
+ * Define Breakpoints here.
+ */
+$mobile-width: 320px;
+$tablet-width: 768px;
+$desktop-width: 1024px;
```
shayan-deriv, line 6: the same here

lib/styles/wallet/styles/devices.scss
aizad-deriv: Will do boss 👍🏼

shayan-deriv: so for here we don't need any changes? 🤔

shayan-deriv: if not let me know to merge it 💯

wojciech-deriv (approved):

PR #101, opened 2024-03-04
Aizad/FEQ-1880/Update Input Component

aizad-deriv:
- Replace values inside of `Input`, `HelperMessage.scss`, `PasswordInput` and `PasswordMeter` with css values (added fallback values).
- Define font size for `Input` component
- Fix minor styling issues
- Update Storybook component for `Input` & `PasswordInput` component.
- Update component with TSDocs.
- Added test cases for `Input` and `HelperMessage`


lib/components/PasswordInput/PasswordInput.scss
```diff
@@ -17,11 +13,11 @@ $ERROR: #ec3f3f;
     &__meter {
         position: absolute;
         z-index: 1;
-        border-radius: 0px 0px 4px 4px;
+        border-radius: 0rem 0rem $border-radius $border-radius;
```
shahzaib-deriv, line 16: why are we using rem here? if we use it in deriv app, the base is 10px so it'll be based on 10px 

if we branch out and we don't set the base to 10px it'll be huge difference

lib/components/Input/Input.scss
```diff
@@ -59,10 +55,10 @@ $border: 1px solid;
         &:focus,
         &:not(:placeholder-shown) {
             & ~ label {
-                font-size: 10px;
+                font-size: 1rem;
```
heorhi-deriv, line 58:
```suggestion
                font-size: 1rem;
```
var(--text-size-xs)?

lib/components/PasswordInput/PasswordInput.scss
aizad-deriv: We are gonna use rem everywhere from now on to match it with `deriv.app`. These style changes were made in a previous PR where I have set the `fontSize` to be equivalent to `62.5%`. You can always redefine the `fontSize` within your body inside your `index.css` as `62.5%` (This will only take affect on `rem` not `px`). 

If you still feel that it's not a good approach feel free to raise it up to the channel and address your concerns. 🙏🏼

heorhi-deriv (approved):

shayan-deriv (approved):

PR #106, opened 2024-03-06
Aizad/FEQ-1566/Added Accordion Component

aizad-deriv:
Created new Accordion component.
Props: 
1. allowMultiple
2. disableAnimation
3. isCompact
4. sections
5. variants

https://github.com/deriv-com/ui/assets/103104395/5df26b79-94ed-44db-a4c6-a3d237d53f60


markwylde-deriv:
Should the whole title row trigger the expand/contract action?

For example, if I click here, should it expand?

Screenshot 2024-03-07 at 07 04 32

aizad-deriv:
> Should the whole title row trigger the expand/contract action?
> 
> For example, if I click here, should it expand?
> 
> Screenshot 2024-03-07 at 07 04 32


Based on production and the design file it shows that only when the user clicks on the Arrow Button only then it will expand:
image

markwylde-deriv:
Is it used for this? It expands/collapses when I click the title.

Screenshot 2024-03-07 at 09 25 38

aizad-deriv:
> Is it used for this? It expands/collapses when I click the title.
> 
> Screenshot 2024-03-07 at 09 25 38

That seems like a single Accordion however in deriv-app it's called a `content-expander`, which has its own special use case and functionality. To be honest you can use the accordion for that but it may lack some of the functionalities of the content-expander.

markwylde-deriv:
> > Is it used for this? It expands/collapses when I click the title.
> > Screenshot 2024-03-07 at 09 25 38
> 
> That seems like a single Accordion however in deriv-app it's called a `content-expander`, which has its own special use case and functionality. To be honest you can use the accordion for that but it may lack some of the functionalities of the content-expander.

Weird. But cool, just wanted to check it's got the same functionality as production, which is seems like it does then. thanks @aizad-deriv

markwylde-deriv (approved):

yaswanth-deriv (approved):

lib/components/Accordion/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+);
+
+export const Accordion = ({
+    allowMultiple = false,
+    disableAnimation = false,
+    isCompact = false,
+    sections,
+    variant,
+}: AccordionProps) => {
+    const [activeIndexes, setActiveIndexes] = useState<number[]>([]);
```
shayan-deriv, line 122: can we also have a prop for the seleted_indexes? 
just like tab component they should be able to control the selected item from their side as well. 
```suggestion
    const [activeIndexes, setActiveIndexes] = useState<number[]>(selectedIndexes || []);
    
        useEffect(() => {
        if (selectedIndexes) {
            setActiveIndexes(selectedIndexes);
        }
    }, [selectedIndexes]);
```

lib/components/Accordion/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+    title: ReactNode;
+    content: ReactNode;
+};
+
+type AccordionSectionProps = {
+    disableAnimation?: boolean;
+    isActiveSection: boolean;
+    isCompact?: boolean;
+    section: AccordionSection;
+    sectionIndex: number;
```
shayan-deriv, line 25: wouldn't it be better to use string ( maybe add an id ) instead of number? just like other components ( check `VerticalTabItems`  )

lib/components/Accordion/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+    useState,
+} from "react";
+import clsx from "clsx";
+import Chevron from "./Chevron.svg";
+import "./Accordion.scss";
+
+type AccordionVariants = "underline" | "bordered" | "shadow";
+
+type AccordionSection = {
+    title: ReactNode;
```
shayan-deriv, line 16:
```suggestion
    id: string;
    title: ReactNode;
```
I'm suggesting this in order to not use index number to detect the active section

lib/components/Accordion/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    disableAnimation={disableAnimation}
+                    isActiveSection={activeIndexes.includes(sectionIndex)}
+                    isCompact={isCompact}
+                    key={sectionIndex}
+                    section={section}
+                    sectionIndex={sectionIndex}
+                    setActiveIndex={() => setActiveIndex(sectionIndex)}
+                    variant={variant}
+                />
+            ))}
```
shayan-deriv, line 152:
```suggestion
            {sections.map((section, sectionIndex) => (
                <AccordionSection
                    disableAnimation={disableAnimation}
                    isActiveSection={activeIndexes.includes(section.id)}
                    isCompact={isCompact}
                    key={sectionIndex}
                    section={section}
                    sectionIndex={section.id}
                    setActiveIndex={() => setActiveIndex(section.id)}
                    variant={variant}
                />
            ))}
```

shayan-deriv (commented):
just wanted to check if it's possible to use string over number for the activeIndex

src/components/Accordion/__test__/Accordion.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+describe("Accordion", () => {
+    it("renders correctly", () => {
+        const { getByText } = render(
+            <Accordion title="Test Title">Test Content</Accordion>,
+        );
+
+        expect(getByText("Test Title")).toBeInTheDocument();
+    });
+
```
shayan-deriv, line 13: can we also add more test cases to test `defaultOpen`, `isCompact`,`title` and `variant` as well?

src/components/Accordion/__test__/Accordion.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+        expect(button).toHaveAttribute("aria-expanded", "false");
+
+        fireEvent.click(button);
+        expect(button).toHaveAttribute("aria-expanded", "true");
+
+        fireEvent.click(button);
+        expect(button).toHaveAttribute("aria-expanded", "false");
+    });
+});
```
shayan-deriv, line 30:
```suggestion
    it("opens and closes when the header is clicked", async () => {
        const { getByRole } = render(
            <Accordion title="Test Title">Test Content</Accordion>,
        );

        screen.debug();
        const button = getByRole("button");

        expect(button).toHaveAttribute("aria-expanded", "false");

        await userEvent.click(button);
        expect(button).toHaveAttribute("aria-expanded", "true");

       await userEvent.click(button);
        expect(button).toHaveAttribute("aria-expanded", "false");
    });
});
```

shayan-deriv: we are using userEvent for all of our tests since it simulates real user interaction

shayan-deriv (approved):

PR #107, opened 2024-03-06
fix: added new classname to modal header

shayan-deriv:
## the issue:
- when no title is passed the close icon inside modal header was left aligned but we expected that it be right

## changes:
- added a new classname and check if there is no children passed to the header, then I'm applying `flex-end` for the `justify-content`

## screenshots:
before: 
image

after:
image

### DIFF
--- a/lib/components/Modal/Modal.scss
+++ b/lib/components/Modal/Modal.scss
@@ -20,9 +20,6 @@
         &--no-border {
             border-bottom: none;
         }
-        &--justify-end{
-            justify-content: flex-end;
-        }
     }
     &__title {
         font-size: 20px;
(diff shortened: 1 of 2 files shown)


markwylde-deriv (approved):

lib/components/Modal/Modal.scss
```diff
@@ -20,6 +20,9 @@
         &--no-border {
             border-bottom: none;
         }
+        &--justify-end{
+            justify-content: flex-end;
```
aizad-deriv, line 24: What you can also do is to define margin-left: auto. This will push the element till the end of the container. 👍🏼 
Let me know if it works 😅

shayan-deriv: yes. it worked. thanks for ur suggestion 👍

amina-deriv (approved):

PR #112, opened 2024-03-07
Aizad/ FEQ-1886 / Add Prefix to Global CSS Variables

aizad-deriv:
- Append prefix `--du` for css variables
- Minor changes inside of stylesheets
- Refactor Text component: Replace hard-coded values with CSS variables


thisyahlen-deriv (approved):

lib/components/Text/index.tsx
```diff
@@ -21,18 +32,11 @@ export interface TextProps {
...
-        | "less-prominent"
-        | "primary"
-        | "success"
-        | "warning";
+    color?: TextColors;
     fontStyle?: CSSProperties["fontStyle"];
     lineHeight?: TGenericSizes;
     size?: Exclude<TGenericSizes, "3xs" | "6xl" | "7xl">;
-    weight?: CSSProperties["fontWeight"];
+    weight?: Exclude<CSSProperties["fontWeight"], "bolder" | "lighter">;
```
shayan-deriv, line 39: Could you please provide more details about this change?

shayan-deriv (approved):

lib/components/Text/index.tsx
aizad-deriv: The only styles we are providing are `bold`, `normal`, `semi-bold` and `light` the rest is not being handled by us. So if the user pass other than the one we specify no changes in the styles will take affect.

PR #125, opened 2024-03-12
fix: fixed text size issue in button component

shayan-deriv:


aizad-deriv (approved):

src/components/Button/index.tsx
```diff
@@ -99,7 +99,7 @@ export const Button = ({
             {rest.children && !isLoading && (
                 <Text
                     align="center"
-                    size={FontSize[size] ?? textSize}
+                    size={ typeof textSize != "undefined" ? textSize : FontSize[size]}
```
ameerul-deriv, line 102: cant we just check for textSize like this? cause it can be undefined right, so we don't need to explicity check typeof, or we can use `!!textSize`
```suggestion
                    size={ textSize ? textSize : FontSize[size]}
```

ameerul-deriv: or actually you can do this, this is better hahaha
```suggestion
                    size={ textSize || FontSize[size] }
```

shayan-deriv: u're right. I've updated it

ameerul-deriv (approved):

PR #146, opened 2024-03-26
[FEQ]Yaswanth/FEQ-1995 Added Test for inline message component

yaswanth-deriv:
Added Test cases for inline message component


src/components/InlineMessage/__test__/InlineMessage.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+    );
+  });
+
+  it("renders with custom class name", () => {
+    const { container } = render(
+      <InlineMessage className="custom-class">Custom Message</InlineMessage>
+    );
+    expect(container.firstChild).toHaveClass("custom-class");
+  });
+});
```
shayan-deriv, line 42: we need another test case for `iconPosition`

shayan-deriv (approved):

PR #148, opened 2024-03-27
[FIX]Yaswanth/To Fix Type errors for modal and dialog component

yaswanth-deriv:
Need to make changes in package.json file to Fix Type errors for modal and dialog component


shayan-deriv:
this is fixed in #152

PR #158, opened 2024-04-01
[FEQ]Yaswanth/FEQ_1913/Improve/Add hoverstyle prop to button

yaswanth-deriv:
Need to add hoverstyle prop to button


src/components/Button/__test__/Button.spec.tsx
```diff
@@ -1,92 +1,109 @@
...
+    expect(screen.queryByText("Test Button")).toBeNull();
   });
+
+  it("hover styles are applied when hasHoverStyle is true", () => {
+    render(
+      <Button hasHoverStyle={true}>Hover Button</Button>
+    );
+    const button = screen.getByRole("button");
+    expect(button).not.toHaveClass("deriv-button__disable-hover");
+  });
```
ameerul-deriv, line 99: since its defaulted to true, you don't need to pass this prop right? could you check 
```suggestion
  it("hover styles are applied when hasHoverStyle is true", () => {
    render(
      <Button>Hover Button</Button>
    );
    const button = screen.getByRole("button");
    expect(button).not.toHaveClass("deriv-button__disable-hover");
  });
```

src/components/Button/index.tsx
```diff
@@ -65,6 +66,7 @@ export const Button = ({
     isLoading = false,
     rounded = "sm",
     size = "md",
+    hasHoverStyle=true,
```
ameerul-deriv, line 69: IMO i think its better to have it set to false first and rename it to hideHoverStyle. This is because in terms of simplicity of when you want to pass this prop to a component it will be a bit more readable and you don't need to pass `false` every time we want to hide the hover styles. For example here are the two differences:

<Button hasHoverStyles={false} /> vs <Button hideHoverStyles />

The second example is more straightforward and will handle setting it to true by simply passing the prop

```suggestion
    hideHoverStyles=false,
```

src/components/Button/index.tsx
```diff
@@ -80,6 +82,7 @@ export const Button = ({
                 ButtonRounded[rounded],
                 {
                     "deriv-button__full-width": isFullWidth,
+                    "deriv-button__disable-hover": !hasHoverStyle,
```
ameerul-deriv, line 85: following my above message: 
```suggestion
                    "deriv-button__disable-hover": hideHoverStyles,
```

src/components/Button/Button.scss
```diff
@@ -196,4 +196,14 @@ $black-outline-hover: rgba(0, 0, 0, 0.08);
             }
         }
     }
+
+    &.deriv-button__disable-hover {
```
shayan-deriv, line 200:
```suggestion
    &.deriv-button__hover--disabled {
```
to follow BEM . also you can move it to line 26 ( before size properties )

src/components/Button/index.tsx
yaswanth-deriv: Yes This is more readable.I will make changes.Thanks

nada-deriv (approved):

src/components/Button/__test__/Button.spec.tsx
```diff
@@ -1,92 +1,109 @@
...
-  it('does not render children text when "isLoading" prop is true', () => {
+  it("does not render children text when isLoading prop is true", () => {
     render(<Button isLoading>Test Button</Button>);
-    expect(screen.queryByText('Test Button')).toBeNull();
+    expect(screen.queryByText("Test Button")).toBeNull();
   });
+
+  it("hover styles are applied when hideHoverStyles is true", () => {
+    render(
+      <Button hideHoverStyles={true}>Hover Button</Button>
```
ameerul-deriv, line 95: nit
```suggestion
      <Button hideHoverStyles>Hover Button</Button>
```

src/components/Button/Button.scss
```diff
@@ -24,6 +24,16 @@ $black-outline-hover: rgba(0, 0, 0, 0.08);
         cursor: not-allowed;
     }
 
+    &.deriv-button__hover--disabled {
```
shayan-deriv, line 27:
```suggestion
    &__hover--disabled {
```
I mean like this

shayan-deriv: @aizad-deriv is this correct? 🤔

shayan-deriv: I man is this following BEM?

yaswanth-deriv: @shayan-deriv I think what you suggested is right.It is working too.

ameerul-deriv (approved):

shayan-deriv (approved):

PR #166, opened 2024-04-18
[FEQ]Yaswanth/FEQ-1765/Improve/Disabled label animation and added props

yaswanth-deriv:
Added "isLabelAnimationDisabled" prop with that we can able to control label animation for input component


src/components/Input/Input.scss
```diff
@@ -81,6 +81,19 @@ $input-placement: 16px;
             }
         }
     }
+    &__fieldDisabled{
```
shayan-deriv, line 84:
```suggestion
    &--label-animation-disabled{
```

shayan-deriv (approved):

PR #168, opened 2024-04-24
[FEQ] / Shayan/ FEQ-2122 / add custom error message to passwordInput component

shayan-deriv:
changes: 
- Added CustomErrorMessage prop to `PasswordInput`
- Added new score value
- Updated the `validatePassword` function 
- Added new story to cover this scenario

### DIFF
--- a/src/components/PasswordInput/PasswordConstants.ts
+++ b/src/components/PasswordInput/PasswordConstants.ts
@@ -1,4 +1,4 @@
-export type TScore = 0 | 1 | 2 | 3 | 4;
+export type TScore = 0 | 1 | 2 | 3 | 4 | 5;
 
 export type passwordKeys =
     | "common"
--- a/src/components/PasswordInput/PasswordMeter.tsx
+++ b/src/components/PasswordInput/PasswordMeter.tsx
@@ -13,6 +13,7 @@ const PasswordStrengthClass: Record<TScore, string> = {
     2: "deriv-password__meter__bar--moderate",
     3: "deriv-password__meter__bar--strong",
     4: "deriv-password__meter__bar--complete",
+    5: "deriv-password__meter__bar--error",
 };
 
 export const PasswordMeter = ({ score }: PasswordMeterProps) => (
--- a/src/components/PasswordInput/PasswordUtils.ts
+++ b/src/components/PasswordInput/PasswordUtils.ts
@@ -31,8 +31,9 @@ export const isPasswordStrong = (password: string) => {
     );
 };
 
-export const calculateScore = (password: string) => {
+export const calculateScore = (password: string, customErrorMessage="") => {
     if (password?.length === 0) return 0;
+    if (customErrorMessage) return 5;
     if (!isPasswordValid(password)) return 1;
     if (
         !isPasswordStrong(password) &&
--- a/src/components/PasswordInput/PasswordUtils.ts
+++ b/src/components/PasswordInput/PasswordUtils.ts
@@ -31,7 +31,7 @@ export const isPasswordStrong = (password: string) => {
     );
 };
 
-export const calculateScore = (password: string, customErrorMessage="") => {
+export const calculateScore = (password: string, customErrorMessage = "") => {
     if (password?.length === 0) return 0;
     if (customErrorMessage) return 5;
     if (!isPasswordValid(password)) return 1;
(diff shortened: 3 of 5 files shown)


src/components/PasswordInput/index.tsx
```diff
@@ -86,19 +94,25 @@ export const PasswordInput = ({
...
+        setBackendErrorMessage(customErrorMessage);
+    }, [customErrorMessage]);
     const [isTouched, setIsTouched] = useState(false);
     const [showPassword, setShowPassword] = useState(false);
+    const [backendErrorMessage, setBackendErrorMessage] = useState(customErrorMessage);
 
     const { errorMessage, score } = useMemo(
-        () => validatePassword(value as string),
-        [value],
+        () => validatePassword(value as string, backendErrorMessage),
```
hasan-deriv, line 108: why `as string`? didn't we already define `value`'s type?

src/components/PasswordInput/PasswordUtils.ts
```diff
@@ -31,8 +31,9 @@ export const isPasswordStrong = (password: string) => {
     );
 };
 
-export const calculateScore = (password: string) => {
+export const calculateScore = (password: string, customErrorMessage="") => {
```
matin-deriv, line 34: [nit]: prettier formatting missed here.

src/components/PasswordInput/index.tsx
```diff
@@ -22,8 +23,8 @@ import { EyeIcon, EyeIconSlash } from "./PasswordIcon";
 import { PasswordMeter } from "./PasswordMeter";
 import "./PasswordInput.scss";
 
-export const validatePassword = (password: string) => {
-    const score = calculateScore(password);
+export const validatePassword = (password: string, customErrorMessage="") => {
```
matin-deriv, line 26: ditto.

matin-deriv (commented):
LGTM, just a nitpick that I left above.

src/components/PasswordInput/PasswordConstants.ts
```diff
@@ -1,4 +1,4 @@
-export type TScore = 0 | 1 | 2 | 3 | 4;
+export type TScore = 0 | 1 | 2 | 3 | 4 | 5;
```
aizad-deriv, line 1: Why do we have another score `5` for custom message? Can we just override the error messages instead?

src/components/PasswordInput/PasswordMeter.tsx
```diff
@@ -13,6 +13,7 @@ const PasswordStrengthClass: Record<TScore, string> = {
     2: "deriv-password__meter__bar--moderate",
     3: "deriv-password__meter__bar--strong",
     4: "deriv-password__meter__bar--complete",
+    5: "deriv-password__meter__bar--error",
```
aizad-deriv, line 16: Is this a new class?

src/components/PasswordInput/index.tsx
```diff
@@ -36,9 +37,14 @@ export const validatePassword = (password: string) => {
         errorMessage = passwordErrorMessage.invalidLength;
     } else if (!isPasswordValid(password)) {
         errorMessage = passwordErrorMessage.missingCharacter;
-    } else {
+    }
+    else if( customErrorMessage){
```
niloofar-deriv, line 41: [nit]: prettier formatting missed here as well :))

src/components/PasswordInput/PasswordConstants.ts
shayan-deriv: we tried but it had conflicts with other cases

src/components/PasswordInput/PasswordMeter.tsx
shayan-deriv: no it's the same as code `0` but I had to add this because of TS error 🤔

src/components/PasswordInput/index.tsx
shayan-deriv: the value can be different based on the input type. but in this case we only need to treat it as `string` . we can add type checking but then we will need to update `validatePassword()` and properly return  `errorMessage`, `score` when they are undefined. it's not in the scope of this card. what do you think? 🤔

matin-deriv (approved):

niloofar-deriv (approved):

src/components/PasswordInput/index.tsx
hasan-deriv: I think we should fix it or create a new card for it. But, it has to be fixed.

thisyahlen-deriv (approved):

src/components/PasswordInput/index.tsx
shayan-deriv: definitely. I'll create a card for it. thanks 🙏

PR #189, opened 2024-05-08
Niloofar/Added menu item component

niloofar-deriv:
This PR introduces the `MenuItem` component, designed to render dynamically as either a `<button>` or an `<a> `tag based on the passed props.
This flexibility allows the component to function in various contexts throughout our application, such as navigating to different pages or performing in-page actions.

https://github.com/deriv-com/ui/assets/93518187/d3e88319-fc75-498c-b9ae-2ca1650a0f6e

Screenshot 2024-05-08 at 6 14 15 PM
Screenshot 2024-05-08 at 6 14 24 PM

### DIFF
--- a/src/components/AppLayout/Header/MenuItem/MenuItem.scss
+++ b/src/components/AppLayout/Header/MenuItem/MenuItem.scss
@@ -1,0 +1,15 @@
+.deriv-menu-item {
+    display: flex;
+    align-items: center;
+    height: 100%;
+
+    &:hover {
+        background: var(--du-general-hover, #e6e9e9);
+    }
+
+    &--active {
+        &:hover {
+            background: transparent;
+        }
+    }
+}
--- a/src/components/AppLayout/Header/MenuItem/MenuItem.scss
+++ b/src/components/AppLayout/Header/MenuItem/MenuItem.scss
@@ -2,6 +2,7 @@
     display: flex;
     align-items: center;
     height: 100%;
+    cursor: pointer;
 
     &:hover {
         background: var(--du-general-hover, #e6e9e9);
--- a/src/components/AppLayout/Header/MenuItem/index.tsx
+++ b/src/components/AppLayout/Header/MenuItem/index.tsx
@@ -1,8 +1,8 @@
 import {
     ComponentProps,
+    ElementType,
     PropsWithChildren,
     ReactNode,
-    createElement,
 } from "react";
 import clsx from "clsx";
 import "./MenuItem.scss";
@@ -16,13 +16,13 @@ import "./MenuItem.scss";
  * @property {boolean} disableHover - If true, disables hover effects.
  * @property {boolean} active - If true, applies an active state style.
  */
-type TMenuItem = ComponentProps<"a" | "button"> & {
-    as: "a" | "button";
+interface TMenuItem extends ComponentProps<ElementType> {
+    as?: "a" | "button";
     leftComponent?: ReactNode;
     rightComponent?: ReactNode;
     disableHover?: boolean;
     active?: boolean;
-};
+}
 
 /**
  * MenuItem component that can render as either an anchor or a button element, with optional left and right components.
@@ -41,7 +41,7 @@ type TMenuItem = ComponentProps<"a" | "button"> & {
  * @returns {React.ReactElement} A React Element of type 'a' or 'button' based on the 'as' prop.
  */
 export const MenuItem = ({
-    as,
+    as = "a",
     leftComponent,
     children,
     rightComponent,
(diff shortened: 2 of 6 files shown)


src/components/AppLayout/Header/MenuItem/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+        ),
+        children: [
+            createElement("div", { key: "leftComponent" }, leftComponent),
+            createElement("div", { key: "mainChildren" }, children),
+            createElement("div", { key: "rightComponent" }, rightComponent),
+        ],
+        ...props,
+    };
+
+    return createElement(as, { ...content });
```
shayan-deriv, line 67:
```suggestion
    const Tag = as;
    return (
        <Tag 
            className={clsx("deriv-menu-item", {
                { "deriv-menu-item--active": active || disableHover },
                className
            )}
            { ...props }
        >
            <div> {leftComponent} </div>
            {children}
            <div> {rightComponent} </div>
        </Tag>
    )
```

src/components/AppLayout/Header/MenuItem/MenuItem.scss
```diff
@@ -0,0 +1,15 @@
+.deriv-menu-item {
+    display: flex;
+    align-items: center;
+    height: 100%;
```
shayan-deriv, line 4: do we need this? 
because without any parents this won't be applied on the element.

src/components/AppLayout/Header/MenuItem/index.tsx
aizad-deriv: Adding a prop as a classname for both left and right sections will be helpful for the developers if they have any specific design requirements:

```
<div className={'menu-left', leftMenuClassName}> {leftComponent} </div>
            {children}
<div className={'menu-right', rightMenuClassName}> {rightComponent} </div>

...
.menu-left {
  display: inline-block;
   margin-right: auto; // this will make sure that the menu will always stick to the left handside of the container 
}

.menu-right {
   display: inline-block;
   margin-left: auto; // this will make sure that the menu will always stick to the right handside of the container
}

// Above are just suggestions 😄  if it is contained within a flexbox then you can add `space-between`.

```

src/components/AppLayout/Header/MenuItem/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+ * @param {ReactNode} props.children - The main content of the MenuItem.
+ * @param {ReactNode} props.rightComponent - Optional component rendered on the right side of the MenuItem.
+ * @param {boolean} props.disableHover - If set to true, no hover effects are applied.
+ * @param {boolean} props.active - If set to true, the 'active' styling is applied.
+ * @param {string} props.className - Additional className for custom styling.
+ * @param {Object} props.otherProps - Spread into the element as additional HTML attributes.
+ * @returns {React.ReactElement} A React Element of type 'a' or 'button' based on the 'as' prop.
+ */
+export const MenuItem = ({
+    as,
```
aizad-deriv, line 44: is there a default value for dat `as`? 🤨

src/components/AppLayout/Header/MenuItem/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+ * @property {boolean} disableHover - If true, disables hover effects.
+ * @property {boolean} active - If true, applies an active state style.
+ */
+type TMenuItem = ComponentProps<"a" | "button"> & {
+    as: "a" | "button";
+    leftComponent?: ReactNode;
+    rightComponent?: ReactNode;
+    disableHover?: boolean;
+    active?: boolean;
+};
```
hasan-deriv, line 25:
```suggestion

// Base Props for MenuItem without the 'as' property
interface BaseMenuItemProps {
    leftComponent?: ReactNode;
    rightComponent?: ReactNode;
    disableHover?: boolean;
    active?: boolean;
}

// Conditional props that extract the correct attributes for 'a' or 'button'
type ConditionalProps<T extends "a" | "button"> = ComponentProps<T> & BaseMenuItemProps & {
    as: T;
};

// TMenuItem type that uses conditional types based on the 'as' property
type TMenuItem = ConditionalProps<"a"> | ConditionalProps<"button">;
```

src/components/AppLayout/Header/MenuItem/index.tsx
shayan-deriv: @aizad-deriv but the leftComponent is an JSX , right? so they can pass it with the className from their side 🤔

src/components/AppLayout/Header/MenuItem/index.tsx
hasan-deriv: I was able to send `href` attribute when `as` property was set to `button`. This type will prevent

hasan-deriv (changes requested):

src/components/AppLayout/Header/MenuItem/index.tsx
niloofar-deriv: @shayan-deriv applied your changes 🙏

niloofar-deriv: @aizad-deriv we don't need to do that the user has full control from his side

src/components/AppLayout/Header/MenuItem/MenuItem.scss
niloofar-deriv: Yes @shayan-deriv it's better to be there cause we are using menuItem in the div and it definitely have a parent so we need this.

src/components/AppLayout/Header/MenuItem/index.tsx
niloofar-deriv: I added this thanks 🙏

src/components/AppLayout/Header/MenuItem/index.tsx
niloofar-deriv: Thanks @hasan-deriv that's a great suggestion but as we talked we have an issue with ref type and since we are going to deliver this component ASAP I'm going to ignore this for now later we can improve the types 🙏

shayan-deriv (approved):

PR #200, opened 2024-05-20
fix: tooltip issue

niloofar-deriv:


src/components/Tooltip/index.tsx
```diff
@@ -1,29 +1,55 @@
...
+const TooltipArrowVariantClass = {
+    dark: "deriv-arrow--dark",
+    error: "deriv-arrow--error",
+    general: "deriv-arrow--general",
+} as const;
+
 export const Tooltip = ({
     children,
     className,
+    activeClassName,
```
shayan-deriv, line 45: what is this  for? 🤔

src/components/Tooltip/index.tsx
```diff
@@ -1,29 +1,55 @@
...
-type TooltipProps = {
-    children?: ReactNode;
-    className?: string;
-    message: ReactNode;
-    position?: TooltipPositionType;
-    triggerAction?: TooltipTriggerActionType;
-    variant?: TooltipVariantType;
+type TTooltip = Omit<
+    ComponentProps<"div">,
+    "ref" | "onMouseEnter" | "onMouseLeave"
```
shayan-deriv, line 14: why removing these? 🤔

shayan-deriv: we should pass all possible props if we are extending the types from div element

niloofar-deriv: Cause we are handling them internally and user dose not need to change these

src/components/Tooltip/index.tsx
niloofar-deriv: For the message wrapper maybe they need to add sth to it

src/components/Tooltip/index.tsx
```diff
@@ -1,29 +1,55 @@
...
     className,
+    activeClassName,
     message,
     position = "top",
     triggerAction = "hover",
     variant = "general",
-}: TooltipProps) => {
+    onClick,
+    ...rest
+}: PropsWithChildren<TTooltip>) => {
```
aizad-deriv, line 52: let's add a forwardref for this component aswell 😃

PR #202, opened 2024-05-21
Niloofar/ TooltipMenuIcon component

niloofar-deriv:
A component that renders a `button` | `a` | `div` with a tooltip.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     },
     "dependencies": {
         "@deriv/quill-icons": "^1.22.5",
-        "@types/react-modal": "^3.16.3"
+        "@types/react-modal": "^3.16.3",
+        "react-tiny-popover": "^8.0.4"
     },
     "devDependencies": {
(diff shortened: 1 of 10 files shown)


ali-hosseini-deriv (approved):

package.json
```diff
@@ -34,7 +34,8 @@
     },
     "dependencies": {
         "@deriv/quill-icons": "^1.22.5",
-        "@types/react-modal": "^3.16.3"
+        "@types/react-modal": "^3.16.3",
```
hasan-deriv, line 37: Shouldn't it be in `devDependencies`?

niloofar-deriv: It is out of the scope of this card, we can change it later 🙏

shayan-deriv: @hasan-deriv we had to put it here because we are including react-modal in our bundle ( since our modal component is build on top of it ) so we needed to include the types as well, otherwise the consumer would have to install it

PR #217, opened 2024-06-12
chore: added className props to the PlatformSwitcher component

niloofar-deriv:


shayan-deriv (approved):

shayan-deriv (changes requested):
could you please add the test case for the new prop as well? 🙏

shayan-deriv (approved):

PR #218, opened 2024-06-14
chore: added new props to platformSwitcher

niloofar-deriv:


shayan-deriv (changes requested):
could u please reflect it inside test and story-book as well? 🙏

shayan-deriv (approved):

PR #222, opened 2024-07-01
Lubega / Fix dropdown UI discrepancies

lubega-deriv:
### Changes:
- [x] Fix small UI discrepancies with dropdown component
- [x] Added right margin for left icon and top margin for dropdown list

## Before:
Screenshot 2024-07-01 at 1 38 20 PM

## After:
Screenshot 2024-07-01 at 1 38 11 PM

### DIFF
--- a/src/components/Dropdown/index.tsx
+++ b/src/components/Dropdown/index.tsx
@@ -160,7 +160,7 @@ export const Dropdown = ({
                 <Input
                     disabled={disabled}
                     label={reactNodeToString(label)}
-                    leftPlaceholder={icon || undefined}
+                    leftPlaceholder={icon}
                     message={isOpen ? " " : errorMessage}
                     name={name}
                     onClickCapture={handleInputClick}
--- a/src/components/Dropdown/index.tsx
+++ b/src/components/Dropdown/index.tsx
@@ -103,7 +103,7 @@ export const Dropdown = ({
         },
         onIsOpenChange({ isOpen }) {
             if (!isOpen && !emptyResultMessage) {
-                clearFilter();  
+                clearFilter();
             }
         },
         onSelectedItemChange({ selectedItem }) {
@@ -159,6 +159,7 @@ export const Dropdown = ({
             <div className="deriv-dropdown__content">
                 <Input
                     disabled={disabled}
+                    hideMessage={isOpen}
                     label={reactNodeToString(label)}
                     leftPlaceholder={icon || undefined}
                     message={errorMessage}
--- a/src/components/Dropdown/index.tsx
+++ b/src/components/Dropdown/index.tsx
@@ -159,10 +159,9 @@ export const Dropdown = ({
             <div className="deriv-dropdown__content">
                 <Input
                     disabled={disabled}
-                    hideMessage={isOpen}
                     label={reactNodeToString(label)}
                     leftPlaceholder={icon || undefined}
-                    message={errorMessage}
+                    message={isOpen ? " " : errorMessage}
                     name={name}
                     onClickCapture={handleInputClick}
                     onKeyUp={() => setShouldFilterList(true)}
--- a/src/components/Input/Input.scss
+++ b/src/components/Input/Input.scss
@@ -156,6 +156,4 @@ $input-placement: 16px;
     padding: 2px 0 0 16px;
     height: 20px;
     width: 100%;
-    display: flex;
-    align-items: center;
 }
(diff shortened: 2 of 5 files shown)


src/components/Input/index.tsx
```diff
@@ -150,16 +154,16 @@ export const Input = forwardRef(
                         </div>
                     )}
                 </div>
+                <div className="deriv-input__helper-message">
```
nada-deriv, line 157: wont moving this up create an empty div space even if we don't have message to be displayed?

src/components/Input/Input.scss
```diff
@@ -152,4 +156,6 @@ $input-placement: 16px;
     padding: 2px 0 0 16px;
     height: 20px;
     width: 100%;
+    display: flex;
+    align-items: center;
```
aizad-deriv, line 160: can we remove this

src/components/Input/index.tsx
lubega-deriv: Yep, but keeping it as it is would cause the components below the dropdown to shift up and down when the dropdown is open/closed due to the missing div. Need advice on which approach to take @nada-deriv @shayan-deriv 

https://github.com/deriv-com/ui/assets/142860499/39dc1946-ebcc-40d4-b955-3736107cb712

src/components/Input/Input.scss
lubega-deriv: Updated

src/components/Input/index.tsx
```diff
@@ -126,7 +126,11 @@ export const Input = forwardRef(
                     )}
                     <input
                         placeholder={label}
-                        className={islabelAnimationDisabled? "deriv-input--label-animation-disabled" : "deriv-input__field" }
+                        className={
+                            islabelAnimationDisabled
+                                ? "deriv-input--label-animation-disabled"
+                                : "deriv-input__field"
```
yashim-deriv, line 132: this would remove the whole `deriv-input__field` instead of just the animation. Is this intended?

src/components/Input/index.tsx
aizad-deriv: Can't you hide the `helperMessage` aswell as the `errorMessage` when the dropdown is open?

lubega-deriv: @aizad-deriv this is already added:
image

src/components/Input/index.tsx
lubega-deriv: @yashim-deriv this is just eslint formatting, the logic is not changed

shayan-deriv: @yashim-deriv yes this in what we needed to do. in some cases we don't want to move the placeholder to top(as label) when user starts typing and instead we want the whole thing gone

src/components/Input/index.tsx
nada-deriv: @lubega-deriv  what about giving an empty div as message from client side, when we want to clear the message? instead of setting hideMessage as true when open? changing the div position in input will affect the app at a lot of other places..

nada-deriv: https://github.com/deriv-com/ui/assets/122768621/b431290e-e0f3-4ab8-9700-63c7f8614110

something like this 
`message={isOpen ? <div></div> :  errorMessage}` from the client side instead.

shayan-deriv: I think we should keep it as it is. for the shifting up and down issue, we can empty the error message when clicki ng on the dropdown button. when the error message is empty string we don't have that issue

lubega-deriv: @nada-deriv @shayan-deriv updated the implementation to display empty string when dropdown is open

src/components/Dropdown/index.tsx
```diff
@@ -161,7 +161,7 @@ export const Dropdown = ({
                     disabled={disabled}
                     label={reactNodeToString(label)}
                     leftPlaceholder={icon || undefined}
```
jia-deriv, line 163: What is the difference when we use `leftPlaceholder={icon}` here instead of `leftPlaceholder={icon || undefined}`?

shayan-deriv: Actually, we don't need the fallback here since the types match.

lubega-deriv: Yep I believe this was previously added as a fallback. Updated and removed this fallback as it's not needed

jia-deriv (approved):

shayan-deriv (approved):

PR #225, opened 2024-07-11
Aizad/FEQ-1874/Dropdown-Component-Fixes

aizad-deriv:
## Issues:
- [x] dropdown arrow rotating animation
- [x] dropdown input field not full width
- [x] empty div issue for helper message placeholder below dropdown
- [x] only the colour of the label dims instead of the whole dropdown when it is disabled

Preview:

https://github.com/deriv-com/ui/assets/103104395/56ab1f30-cb2a-4337-9350-5ca314e08556

### DIFF
--- a/src/components/Dropdown/index.tsx
+++ b/src/components/Dropdown/index.tsx
@@ -139,6 +139,7 @@ export const Dropdown = ({
         <div
             className={clsx("deriv-dropdown", {
                 "deriv-dropdown--disabled": disabled,
+                "deriv-dropdown--full": isFullWidth,
             })}
             {...getToggleButtonProps()}
         >
@@ -169,6 +170,7 @@ export const Dropdown = ({
                     }
                     type="text"
                     value={value}
+                    isFullWidth={isFullWidth}
                     {...getInputProps()}
                     {...rest}
                 />
--- a/src/components/Dropdown/index.tsx
+++ b/src/components/Dropdown/index.tsx
@@ -7,15 +7,15 @@ import React, {
 } from "react";
 import clsx from "clsx";
 import { useCombobox } from "downshift";
-import "./Dropdown.scss";
+import { LegacyChevronDown2pxIcon } from "@deriv/quill-icons";
 import { Input } from "../Input";
 import { TGenericSizes } from "../../types";
 import { Text } from "../Text";
+import "./Dropdown.scss";
 
 type InputProps = React.ComponentProps<typeof Input>;
 type TProps = HtmlHTMLAttributes<HTMLInputElement> & {
     disabled?: boolean;
-    dropdownIcon: React.ReactNode;
     emptyResultMessage?: string;
     errorMessage?: InputProps["message"];
     icon?: React.ReactNode;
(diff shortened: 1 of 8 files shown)


src/components/Dropdown/index.tsx
```diff
@@ -166,9 +153,24 @@ export const Dropdown = ({
...
+                                    "deriv-dropdown__chevron--open": isOpen,
+                                    "deriv-dropdown__chevron--disabled":
+                                        disabled,
+                                })}
+                                iconSize="xs"
+                            />
+                        </button>
+                    }
                     type="text"
                     value={value}
```
jia-deriv, line 172: If the value is not a state when we pass in `Option1` from the props will the dropdown active item will be always `Option1` after we click on other option in the dropdown list? For example if we click on `Option2` then the active item will still be `Option1`.

aizad-deriv: @jia-deriv Hey thanks for spotting that I'll make the amendments right now.

shayan-deriv: @jia-deriv, that scenario won't occur because the value prop is used in line 86 to set the default selected item:

```
    } = useCombobox({
        defaultSelectedItem: items.find((item) => item.value === value) ?? null,
``` 

This ensures that the default value is handled internally by the useCombobox hook from the Downshift library. When the user selects a different option, Downshift will manage the change, so there's no need to worry about the active item reverting to Option1.

@aizad-deriv correct me if I'm wrong 🙏

jia-deriv: @shayan-deriv yes the input value will handle by Downshift we don't have to worry about it. But for the dropdown item has different color for active item.

```typescript
 isOpen &&
        (items.length ?
            items.map((item, index) => ( <
                    li className = {
                        clsx("deriv-dropdown__item", {
                            "deriv-dropdown__item--active": value === item.value // the value will take it from the prop
                        })
                    }
```

In this case the if we passing "Option1" as value in the props then "Option1" will always active in the dropdown item but the input value will change.

aizad-deriv: What @shayan-deriv said is true. It is being handled by downshift. I think from storybook you can see that it is not being updated because it is passed as string instead of state.

jia-deriv: In this case maybe we can use `selectedItem` returned prop from `useCombobox` to style here. Since the value handle by downshift.

jia-deriv: Yea, the state is better to managing from consumer side. Thanks @aizad-deriv @shayan-deriv

aizad-deriv: it is passed as a state for the so it gets changed as expected. We are using this `uncontrolled` component practice towards the whole package. You can refer below how the P2P squad is doing it:
```
<Dropdown
                dropdownIcon={<LabelPairedChevronDownMdRegularIcon />}
                label={localize('Sort by')}
                list={list as unknown as MutableOption[]}
                name='Sort by'
                onSelect={value => onSelect(value as TSortByValues)}
                value={value}
            />
```

jia-deriv (approved):

shayan-deriv (approved):

PR #234, opened 2024-08-08
Lubega/ Fix button ghost hover effect

lubega-deriv:
### Changes:
- [x] Fix hover effect for ghost button to match core components design 

### NEW vs OLD:
https://github.com/user-attachments/assets/f5a86601-d60d-41c5-b333-5367f5b17133

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -141,10 +141,7 @@ $black-outline-hover: rgba(0, 0, 0, 0.08);
                     color: $primary;
                 }
                 &:hover:not(:disabled) {
-                    background-color: #eb3e48;
-                    span {
-                        color: $white;
-                    }
+                    background-color: #ff444f14;
                 }
             }
             &.deriv-button__color--primary-light {
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -7,6 +7,8 @@ $primary-hover: #eb3e48;
 $primary-light-hover: #ff444f3d;
 $white-hover: #d6dadb;
 $black-hover: #323738;
+$primary-outline-hover: #ff444f14;
+$primary-light-outline-hover: #ff444f3d;
 $white-outline-hover: #ffffff14;
 $black-outline-hover: rgba(0, 0, 0, 0.08);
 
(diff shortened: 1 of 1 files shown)


jim-deriv (approved):

nada-deriv:
@lubega-deriv  isn't the variant already correct ? 

https://github.com/user-attachments/assets/b602783b-0ffc-40bc-930f-4eb180797edf

src/components/Button/Button.scss
```diff
@@ -141,26 +143,23 @@ $black-outline-hover: rgba(0, 0, 0, 0.08);
                     color: $primary;
                 }
                 &:hover:not(:disabled) {
-                    background-color: #eb3e48;
-                    span {
-                        color: $white;
-                    }
+                    background-color: $primary-outline-hover;
```
farrah-deriv, line 146: why are we changing this?

ameerul-deriv: IMO its better to add another hover variant for this option, we do have one button with the hover effect in our Header

lubega-deriv: @farrah-deriv @ameerul-deriv currently this `primary` ghost button with bright red hover effect below is not in use anywhere and does not match the figma design for core components, which is why the colour is changed similar to the `primary-light` ghost button. But if we want to keep it as separate variants we can follow the current implementation and close this PR

https://github.com/user-attachments/assets/a66ebc86-fe15-4402-b61a-753fac0e010e

ameerul-deriv: you could still add another variant for it, cause i think the one you want is similar to production right? the v2 is slightly different:

V2:
Screenshot 2024-08-08 at 11 16 18 AM

V1:
Screenshot 2024-08-08 at 2 00 34 PM

ameerul-deriv: but we gotta have some alignment across the different design teams 😆

lubega-deriv: @ameerul-deriv  Seems like we already have this variant, which is called Primary Transparent. So for now we can use this variant in our implementation and close this PR 🙇 
image

image

PR #239, opened 2024-08-21
fix: add dropdown icon prop

nada-deriv:
Currently the dropdown component doesn't accept any prop for dropdown icon and takes the chevron as default. But there are cases where search icon / ellipsis is needed for the dropdowns. This PR adds a dropdownIcon prop to the component, so if the icon is passed, the passed icon will be taken, or else it defaults to chevron

Screenshot 2024-08-21 at 10 26 19 AM

### DIFF
--- a/src/components/Dropdown/index.tsx
+++ b/src/components/Dropdown/index.tsx
@@ -15,8 +15,8 @@ import "./Dropdown.scss";
 
 type InputProps = React.ComponentProps<typeof Input>;
 type TProps = HtmlHTMLAttributes<HTMLInputElement> & {
+    chevronIcon?: React.ReactNode;
     disabled?: boolean;
-    dropdownIcon?: React.ReactNode;
     emptyResultMessage?: string;
     errorMessage?: InputProps["message"];
     icon?: React.ReactNode;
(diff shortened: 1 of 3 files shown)


aizad-deriv (approved):

src/components/Dropdown/index.tsx
```diff
@@ -37,6 +38,7 @@ type TProps = HtmlHTMLAttributes<HTMLInputElement> & {
 
 export const Dropdown = ({
     disabled,
+    dropdownIcon,
```
farrah-deriv, line 41: what's the difference between `dropdownIcon` and `icon`?

nada-deriv: that's for the left placeholder like currency.. this is for the actual dropdown icon

shayan-deriv, line 41:
```suggestion
    arrowIcon,
```
can we rename it to this so it will be more clear which icon we mean?

shayan-deriv: or `chevronIcon`

stories/Dropdown.stories.ts
```diff
@@ -7,7 +7,6 @@ const meta = {
     tags: ["autodocs"],
     args: {
         disabled: false,
-        dropdownIcon: "",
         errorMessage: "Error message",
```
shayan-deriv, line 10:
```suggestion
        chevronIcon: "",
        errorMessage: "Error message",
```

shayan-deriv (approved):

stories/Dropdown.stories.ts
nada-deriv: will add another story for passed chevronicon instead.

PR #246, opened 2024-09-02
maryia/refactor: improve types of notifications component

maryia-matskevich-deriv:
On the bot side, we don't use the `loadMoreFunction` property of the `Notifications` component. 
Let's make it optional instead of passing an empty function, which causes the eslint to fail.

### DIFF
--- a/src/components/AppLayout/Notifications/index.tsx
+++ b/src/components/AppLayout/Notifications/index.tsx
@@ -20,7 +20,7 @@ export const Notifications = ({
     componentConfig,
     className,
     excludedClickOutsideClass = '',
-    loadMoreFunction,
+    loadMoreFunction = () => { },
     isLoading,
     ...rest
 }: Omit<TNotificationsProps, "style">) => {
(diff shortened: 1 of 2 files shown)


src/components/AppLayout/Notifications/Notification.tsx
```diff
@@ -8,7 +8,7 @@ export const Notification = ({
     buttonAction,
     actionText,
 }: TNotificationObject) => {
-    const haveAction = typeof buttonAction === 'function' && actionText;
+    const haveAction = typeof buttonAction === 'function' && typeof actionText === 'string';
```
maryia-matskevich-deriv, line 11: `buttonAction` can have an empty string as a text but still should work. On the bot side, we do not have a button for the announcement. We have an item that works like a button.

maryia-matskevich-deriv: without this change, the functionality of announcements is broken on the bot side😢

shayan-deriv (approved):

PR #258, opened 2024-09-30
Ameerul / Add Scroll To Accordion Functionality

ameerul-deriv:
- Currently, the accordion component is not able to be scrolled to by whoever consumes it. Now users can pass a ref to the accordion, and a function to the component.
- The new function `onScrollToAccordion` which is optional, and it should be handled by the consumer as for my case, I needed the accordion to be scrolled to the top of the page, and this needs to be handled by the main div that surrounds all of my accordions.
- Also I changed the overflow value to scroll as using auto will add the scrollbar to the content as it opens, which causes it to shift the content. Scroll will already add it by default, and won't cause a shift.

### Scrollbar Content Shift Issue:

https://github.com/user-attachments/assets/ebf7a187-08ff-41e4-ad33-988c1edd8c81

### Scrollbar Content Shift Fix:

https://github.com/user-attachments/assets/fa536280-7920-4d02-be18-e782b5f9e6ee

### Accordion Scroll To:

https://github.com/user-attachments/assets/673e5f23-3eb5-45b0-9057-d185b732e0f0

### DIFF
--- a/src/components/Accordion/Accordion.scss
+++ b/src/components/Accordion/Accordion.scss
@@ -66,7 +66,7 @@ $animation-duration: 0.3s;
     }
     &__content {
         width: 100%;
-        overflow: auto;
+        overflow: scroll;
         opacity: 0;
         background-color: var(--content-bg-color);
         transition: all $animation-duration ease;
--- a/src/components/Accordion/index.tsx
+++ b/src/components/Accordion/index.tsx
@@ -4,6 +4,7 @@ import React, {
     ReactNode,
     useEffect,
     ComponentProps,
+    forwardRef,
 } from "react";
 import Chevron from "./Chevron.svg";
 import clsx from "clsx";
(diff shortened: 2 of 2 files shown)


aizad-deriv (approved):

farrah-deriv (commented):
Can you please check the arrow direction when the accordion is opened? It should be pointing downward, right?

ameerul-deriv:
@farrah-deriv when its opened it should be pointing up when its open, this is how it was already in staging

Screenshot 2024-09-30 at 1 07 03 PM

nada-deriv:
@ameerul-deriv  in the video, the accordion shows pointing to the left

ameerul-deriv:
@nada-deriv @farrah-deriv i forgot to remove the changes to the arrow since i was testing something before 😅 but i assure you its not pushed in the changes. 

In my local, the reason why its rotated left was cause i changed the transform value here, which you can check in code changes, i didn't push


Screenshot 2024-09-30 at 2 17 51 PM

Here is a short demo:


https://github.com/user-attachments/assets/673e5f23-3eb5-45b0-9057-d185b732e0f0

shayan-deriv (approved):

PR #259, opened 2024-10-02
[FEQ] george / FEQ-2683 / Implement Breadcrumbs Component

heorhi-deriv:
- Implement Breadcrumbs Component
- Add test for Breadcrumbs Component
- Add story for Breadcrumbs Component
Screenshot 2024-10-02 at 17 30 29

### DIFF
--- a/src/components/Breadcrumbs/index.tsx
+++ b/src/components/Breadcrumbs/index.tsx
@@ -1,0 +1,55 @@
+import { ComponentProps, ReactNode } from "react";
+import clsx from "clsx";
+import { LegacyChevronRight1pxIcon } from "@deriv/quill-icons";
+import { Text } from "../Text";
+import "./Breadcrumbs.scss";
+
+type TItem = {
+    value: string | number;
+    text: string | JSX.Element;
+};
+
+type BreadcrumbsProps = {
+    className?: string;
+    items: TItem[];
+    handleOnClick: (item: TItem) => void;
+    separator?: ReactNode;
+    textSize?: ComponentProps<typeof Text>["size"];
+};
+
+export const Breadcrumbs = ({
+    className,
+    items,
+    handleOnClick,
+    separator,
+    textSize = "sm",
+}: BreadcrumbsProps) => {
+    return (
+        <ul className={clsx("deriv-breadcrumbs", className)}>
+            {items.map((item: TItem, idx: number) => {
+                const isLastItem = idx === items.length - 1;
+
+                return (
+                    <li key={item.value} className="deriv-breadcrumbs__item">
+                        <Text
+                            className={clsx("deriv-breadcrumbs__text", {
+                                "deriv-breadcrumbs__text--active": isLastItem,
+                            })}
+                            size={textSize}
+                            onClick={() => handleOnClick(item)}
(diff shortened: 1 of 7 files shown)


src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,55 @@
+import { ComponentProps, ReactNode } from "react";
+import clsx from "clsx";
+import { LegacyChevronRight1pxIcon } from "@deriv/quill-icons";
+import { Text } from "../Text";
+import "./Breadcrumbs.scss";
+
+type TItem = {
+    value: string | number;
+    text: string | JSX.Element;
+};
```
shayan-deriv, line 10: just a suggestion. 

```suggestion
type TItem = {
    text: string | JSX.Element;
};
```

and then have a function for generating the key:

```
const getKey = (content: React.ReactNode): string => {
  const extractedText = reactNodeToString(content);
  return extractedText || `item-${Math.random().toString(36).substr(2, 9)}`;
};

const reactNodeToString = function (reactNode: React.ReactNode): string {
        let string = "";
        if (typeof reactNode === "string") {
            string = reactNode;
        } else if (typeof reactNode === "number") {
            string = reactNode.toString();
        } else if (reactNode instanceof Array) {
            reactNode.forEach(function (child) {
                string += reactNodeToString(child);
            });
        } else if (isValidElement(reactNode)) {
            string += reactNodeToString(reactNode.props.children);
        }
        return string;
    };
```
and then use it as the key:
```
<li key={getKey(item.content)}... />
```

lubega-deriv: @shayan-deriv wouldn't generating keys using Math.random() cause unpredictable behaviour and lead to rerendering issues? If the concern is having the key to be unique for the specific component maybe it can be something like this instead:
```
key={`deriv-breadcrumb-${item.value}`}
```

Stackoverflow reference: https://stackoverflow.com/questions/29808636/when-giving-unique-keys-to-components-is-it-okay-to-use-math-random-for-gener

heorhi-deriv: hey @shayan-deriv, not sure that generating keys on the fly is a good idea, actually this is an anti-pattern which can lead to performance issues.

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+                        <Text
+                            className={clsx("deriv-breadcrumbs__text", {
+                                "deriv-breadcrumbs__text--active": isLastItem,
+                            })}
+                            size={textSize}
+                            onClick={() => handleOnClick(item)}
+                        >
+                            {item.text}
+                        </Text>
+                        {idx < items.length - 1 &&
```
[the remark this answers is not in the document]
heorhi-deriv: sorted

src/components/Breadcrumbs/index.tsx
shayan-deriv: @heorhi-deriv @lubega-deriv you're absolutely right. I didn't think about that point. thanks guys

shayan-deriv (approved):

PR #260, opened 2024-10-03
[FIX] Separator visibility

heorhi-deriv:
1. When we are passing separator prop as `separator={is_rtl && <LegacyChevronLeft1pxIcon />}` nothing is rendering because of `??` operator.
Fix it using `||` operator instead.
2. Minor fix for hover state and style
3. Add RTL mixin


shayan-deriv (approved):

shayan-deriv:
LGTM

PR #267, opened 2024-11-25
fix: account switcher design

mayuran-deriv:


shayan-deriv (commented):
@mayuran-deriv May I know the reason for these changes? 
it will impact other apps which use it

src/components/Modal/index.tsx
```diff
@@ -1,27 +1,27 @@
...
-        borderRadius: 0,
+        borderRadius: '10px',
         bottom: 'auto',
         left: '50%',
         margin: 0,
         marginRight: '-50%',
         padding: 0,
         right: 'auto',
-        top: '50%',
+        top: '20%',
```
shayan-deriv, line 21: this will break other apps since they use modal a lot specially in p2p. instead you should pass `styles` as prop to `Modal` to override those basic styles. can do it llke this from `src/components/AppLayout/AccountSwitcher/index.tsx` line 105:

```
                        <Modal
                            closeTimeoutMS={200}
                            onRequestClose={(e) => {
                                e.stopPropagation();
                                closeAccountSwitcher();
                            }}
                            className="deriv-account-switcher__container--mobile"
                            isOpen={isOpen}
                            style={{
                                content: {
                                   top: '20%',
                                   borderRadius: '10px',
                                }
                            }}
                        >
```

shayan-deriv (changes requested):

